PR #6, opened 2018-08-20
account for definition mismatch on aarch64 android

nivkner:
The `libc` crate is using `c_uint` for `sa_flags` on 64 bit and `ulong` on 32 bit when the os is Android.
See the definition [here](https://github.com/rust-lang/libc/blob/master/src/unix/notbsd/android/b64/mod.rs). 
Compiled on a 64 bit Android and the tests passed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -184,11 +184,17 @@ impl Slot {
         // C data structure, expected to be zeroed out.
         let mut new: libc::sigaction = unsafe { mem::zeroed() };
         new.sa_sigaction = handler as usize;
-        #[cfg(target_os = "android")]
+        #[cfg(all(target_os = "android", target_pointer_width = "32"))]
         fn flags() -> libc::c_ulong {
             (libc::SA_RESTART as libc::c_ulong)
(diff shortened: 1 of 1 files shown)


vorner:
Hmm, but it unfortunately fails tests on anything else :-(. Anyway, having *third* version of the `flags` function feels like a bit too much. I've tried [something else](https://github.com/vorner/signal-hook/commit/f464a908025019443296827607270756e69e24a7) ‒ letting the compiler auto-convert the types to a common one.

Could you try out the [auto-flags](https://github.com/vorner/signal-hook/tree/auto-flags) branch? Does it work for you? If so I'd release a new version.

Also, if you by any chance had a way to set up the tests for android on travis (or anything else, for that matter), I would be happy to include them in the CI.

nivkner:
Thank you, you solution is better.
It doesn't work on aarch64 Android as is, because `SA_SIGINFO` is also defined differently then the rest, but once it is adjusted with the same technique, it works.

vorner:
Just released in 0.1.4.

PR #8, opened 2018-11-28
Expose the type of signal numbers as SigNo

jmmv:
Before this change, users of signal-hook might have had to directly
reference libc::c_int in order to be agnostic to the size of the
integers used to represent signal numbers.  This is unfortunate.

With the advent of nix, it's reasonable for programs to stay away
from any direct use of libc -- and it'd be nice if using signal-hook
didn't require the reintroduction of a direct dependency on libc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
+# Upcoming (unreleased) changes
+
+* Exposed a new `SigNo` type to represent the correct integer type for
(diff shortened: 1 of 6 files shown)


jmmv:
Actually... let me send you a different PR with a higher abstraction than this. We can discuss there which one is better -- will keep the two commits around.

PR #9, opened 2018-11-28
Turn signal numbers into an enum

jmmv:
This PR contains two changes: one to expose the raw integer type of the signals to avoid users having to depend on `libc::c_int`, and another change to turn the signal numbers into an enum. The latter follows the pattern of the `nix` crate for handling errnos and the new "feature" is that now signals can be formatted for printing in a nicer manner.

These two changes don't really make sense as separate commits but I'm keeping them as such for now because they are two separate ideas. If you agree to this proposal, I can fold them and adjust the description and `CHANGELOG.md` entry to make more sense as a single unit. But if you think this is a bad idea, I can drop the latter commit and return to the original PR https://github.com/vorner/signal-hook/pull/8 that only included the first commit.

Let me know! Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Upcoming (unreleased) changes
 
-* Exposed a new `SigNo` type to represent the correct integer type for
+* Exposed a new `SigNoInt` type to represent the correct integer type for
   signal numbers.  This prevents users of this library from having to
-  directly address `libc::c_int`.
+  directly address `libc::c_int` or know that they are `i32`s.
+
+* Exposed signal numbers as a `SigNo` enum with the ability to format their
+  values as user-friendly strings.  Previous implementations exposed these
+  as raw `i32`s but the internal representation of `SigNo` is the same so
+  old code should continue to compile unmodified.
 
 # 0.1.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
+# Upcoming (unreleased) changes
+
+* Exposed a new `SigNo` type to represent the correct integer type for
+  signal numbers.  This prevents users of this library from having to
+  directly address `libc::c_int`.
+
 # 0.1.6
 
 * The internally used ArcSwap thing doesn't block other ArcSwaps now (has
   independent generation lock).
+
 # 0.1.5
 
 * Re-exported signal constants, so users no longer need libc.
--- a/src/flag.rs
+++ b/src/flag.rs
@@ -135,7 +135,7 @@ use std::io::Error;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Arc;
 
-use {SigId, SigNo};
+use {SigId, SigNo, SigNoInt};
 
 /// Registers an action to set the flag to `true` whenever the given signal arrives.
 pub fn register(signal: SigNo, flag: Arc<AtomicBool>) -> Result<SigId, Error> {
@@ -163,7 +163,7 @@ mod tests {
     fn self_signal() {
         unsafe {
             let pid = libc::getpid();
-            libc::kill(pid, ::SIGUSR1);
+            libc::kill(pid, SigNo::SIGUSR1 as SigNoInt);
         }
     }
 
@@ -185,7 +185,7 @@ mod tests {
     fn register_unregister() {
         // When we register the action, it is active.
         let flag = Arc::new(AtomicBool::new(false));
-        let signal = register(::SIGUSR1, Arc::clone(&flag)).unwrap();
+        let signal = register(SigNo::SIGUSR1, Arc::clone(&flag)).unwrap();
         self_signal();
         assert!(wait_flag(&flag));
         // But stops working after it is unregistered.
--- a/src/flag.rs
+++ b/src/flag.rs
@@ -135,12 +135,10 @@ use std::io::Error;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Arc;
 
-use libc::c_int;
-
-use SigId;
+use {SigId, SigNo};
 
 /// Registers an action to set the flag to `true` whenever the given signal arrives.
-pub fn register(signal: c_int, flag: Arc<AtomicBool>) -> Result<SigId, Error> {
+pub fn register(signal: SigNo, flag: Arc<AtomicBool>) -> Result<SigId, Error> {
     // We use SeqCst for two reasons:
     // * Signals should not come very often, so the performance does not really matter.
     // * We promise the order of actions, but setting different atomics with Relaxed or similar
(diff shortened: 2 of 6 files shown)


vorner:
Hello

The type alias looks beneficial. My only worry there is probably the name ‒ the C api uses either `sig` or `signum` and it would be nice to be consistent with one of these.

I'm a bit wary about the enum approach, though. I see several disadvantages:
* I don't think there's any guarantee that `c_int` is actually `i32`. Aren't there platforms where these are not the same? Or, are we sure there never will be?
* Also, some platforms might support additional signals. The enum prevents them from being used legally with this library. I'm sure the list of re-exported constants here is not exhaustive.
* The ability to represent an unknown signal/-1 in the enum kind of beats the whole purpose of enums to only support legal variants.
* It's a semver-breaking change. That's not a big problem, but it is a bit annoying anyway. (And I think it could be dealt with in some compatible way. If not, I'll have to do the trick where the last release in a line depends on the higher line.)

But I see some good in having a separate type for the signals ‒ like being able to print them not just as numbers. Could we explore alternative ways to see if some might look better? Some not fully formed ideas:

* Drop the `repr(i32)` attribute and add an `Other(c_int)` variant. Then turn the accepting functions to take `S: Into<SigNum>` (or the other way, `S: Into<c_int>`) and provide `From` and `Into` between the two.
* Link to the [Signal](https://docs.rs/nix/0.11.0/nix/sys/signal/enum.Signal.html) enum in nix and make it somehow compatible. I don't want to introduce a hard dependency on nix (I know some folks don't like it, and it is kind of big chunk of code anyway) and I would like this library to stay slim. But either putting it under a feature or sending a MR into nix to add the `desc` method and `Into<c_int>` to it might work.
* A newtype `SigNum(pub c_int)` with methods and `Into/From` conversions?

jmmv:
Hey, sorry. While I'd like to see this improved in any of the ways you suggested... I don't have the time right now to continue this PR. Feel free to handle as you best see fit. Thanks!

vorner:
OK, thanks for the info. It'll probably take some time before I get around to it either, though.

igalic:
let's see how much time i have to offer: after first using nix' signals, i'd really like to see this PR in signal_hook.

i'm currently working on adding signal handling to a programme i use (https://github.com/void-rs/void/pull/44) and i feel like signal_hook would make for a much better abstractions than nix

vorner:
Sorry, it never was a high priority for me so I haven't gotten around to it yet. If you have the time to do it, it would be great.

igalic:
@vorner i'd like to respond to some of your comments, by essentially pointing at how `nix` is doing it:

> - Drop the `repr(i32)` attribute and add an `Other(c_int)` variant. Then turn the accepting functions to take `S: Into<SigNum>` (or the other way, `S: Into<c_int>`) and provide `From` and `Into` between the two.

```rust
libc_enum!{
    // Currently there is only one definition of c_int in libc, as well as only one
    // type for signal constants.
    // We would prefer to use the libc::c_int alias in the repr attribute. Unfortunately
    // this is not (yet) possible.
    #[repr(i32)]
    pub enum Signal {
```

> - Also, some platforms might support additional signals. The enum prevents them from being used legally with this library. I'm sure the list of re-exported constants here is not exhaustive.
> - The ability to represent an unknown signal/-1 in the enum kind of beats the whole purpose of enums to only support legal variants.

`nix`' `Signal` is quite extensive, and covers all signals that currently exist on all Unix flavours. … by which i mean: Linux, osx, *bsd, android, and emscripten(?!?!)

> - Link to the Signal enum in nix and make it somehow compatible. I don't want to introduce a hard dependency on nix (I know some folks don't like it, and it is kind of big chunk of code anyway) and I would like this library to stay slim.

The two libraries are license compatible.
it would be super neat if we could somehow reuse their work.

if we only `use nix::sys::signal::Signal::{self, *};`, will the rust compiler (linker?) shake the size down to only the bits used?

vorner:
* The i32 vs. c_int point

The fact someone (nix) is doing something doesn't make it necessarily *right*. While it might be true that *today* the only c_int in there is this one, this doesn't have to be the case *forever*. Nix is relying on something the standard *doesn't* promise. One of my goals here is to be correct according to whatever standards exist, not only according to existing implementations.

And, grepping through libc code it seems their developers are *ready*  for introducing a platform that has a different `c_int` than `i32` ‒ otherwise it wouldn't be part of every platform file separately:

```
src/sgx.rs:16:pub type c_int = i32;
src/switch.rs:16:pub type c_int = i32;
src/wasi.rs:5:pub type c_int = i32;
src/cloudabi/mod.rs:14:pub type c_int = i32;
src/hermit/mod.rs:29:pub type c_int = i32;
src/redox/mod.rs:14:pub type c_int = i32;
src/fuchsia/mod.rs:21:pub type c_int = i32;
src/unix/mod.rs:19:pub type c_int = i32;
src/windows/mod.rs:16:pub type c_int = i32;
```

> nix' Signal is quite extensive, and covers all signals that currently exist on all Unix flavours.

Again, that's the case of supporting everything existing *today*. Tomorrow, a new system can appear that has some more signals. Or someone of the linux kernel developers may decide to add one more interesting signal. I wouldn't be surprised if some special-purpose device had some (unpublished) patches adding signals for its own operation (connected fusion reactor overheating, asking applications to do only strictly necessary stuff right now, that by default stops the application but unlike SIGSTOP can be overridden, for example) that exist only on that platform. I'd like to be able to support these even if not necessarily with the same comfort as the „normal“ signals.

> if we only use nix::sys::signal::Signal::{self, *};, will the rust compiler (linker?) shake the size down to only the bits used?

It'll likely be stripped away during link time. But you'll still have to let rustc compile it in the first place. It is another dependency in the dep-tree which someone might need to read and audit for problems before being allowed to use it.

So I'd still prefer to offer compatibility with nix only as a feature-flag. I'm OK with providing From/Into traits to convert or whatever. But nix itself feels like too big chunk to pull in unconditionally just to have one enum with named signals. Especially if there are valid (yet probably rare) reasons to be able to use something not part of the enum.

I respect the choices the nix library made, but I believe they don't fit into *this* library that well.

vorner:
I've decided to split the crate into the low-level part (register/unregister) and the convenient APIs and patterns on top of that. I expect the API of the low-level part (signal-hook-registry) to be stable and not change. The split however allows for easier breaking changes in the higher level. So, if there's a good reason to make a breaking change in the (high-level) API, it is now possible.

Additionally, it also allows wrapping the registration into a *different* higher level API while still sharing the „global“ registry of actions (eg. allowing to have multiple actions on one signal). So if someone wants a crate that works with enums instead of `c_int` signals, it should be possible to build it on top of either the low-level or the high-level part (whatever makes sense).

vorner:
This is not moving forward and the code is stale, I'm going to close this PR. If someone wants to create a separate crate depending on singnal-hook-registry, or try submitting a PR that doesn't have this many drawbacks, they are welcome.

PR #12, opened 2019-03-31
Fix typo in doc comment

foxbenjaminfox:
"the hot path of moth programs" should instead be "the hot path of most programs", I think.


vorner:
Yes, of course you're right. Thank you :-).

I think this can wait to be released with whatever new feature or fix comes next, not doing a separate release just for this.

PR #18, opened 2019-07-15
Windows support

qnighy:
I've tried implementing signal-hook on Windows with almost equivalent API.

## Difference in behavior

- Only `SIGINT` and `SIGTERM` are emulated through `CTRL_C_EVENT` and `CTRL_BREAK_EVENT`.
- Not all Ctrl-C operations work. As far as I've tested in a small example, it works if the program is launched from classical terminal + cmd. It doesn't capture Ctrl-C if the program is launched from mintty + msys2.
- By `SetConsoleCtrlHandler`'s nature, the handlers are called in a separate thread.
- If other handlers are already registered, they will be skipped.

## Difference in API

- Only a part of signal constants is provided from Windows CRT. Other signal constants are manufactured by `(Number in Linux x86) + 64`.
- The emulation layer is at `signal-hook`, not `signal-hook-registry`. No `register_unchecked` provided.
- `signal_hook::pipe` depends on `std::os::unix`, which means it's inherently `#[cfg(unix)]` only. I implemented a similar module called `signal_hook::pipe_windows`.
- ~~There seems no way to generate the Ctrl-C/Ctrl-Break behavior from programs, like `libc::kill`. I exposed `signal_hook::__emulate_kill` which emulates `kill` behavior for testing.~~ <br>
  I've found [`GenerateConsoleCtrlEvent`](https://docs.microsoft.com/ja-jp/windows/console/generateconsolectrlevent) later. I'll modify this PR to use the function.

## Difference in implementation

- We don't need async-signal-safety due to `SetConsoleCtrlHandler` launching a separate thread for handlers. Therefore, in `signal_hook::iterator`, I implemented the same functionality using a usual synchronization primitives (Mutex + Condvar). There seems no handy portable equivalent of `UnixSocket`.

### DIFF
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -4,6 +4,7 @@
 )]
 #![deny(missing_docs, warnings)]
 #![allow(unknown_lints, bare_trait_objects)]
+#![cfg(not(windows))]
 
 //! Backend of the [signal-hook] crate.
 //!
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -25,6 +25,9 @@
 //! #       signal_hook::SIGUSR1,
 //!     ])?;
 //! #   // A trick to terminate the example when run as doc-test. Not part of the real code.
+//! #   #[cfg(windows)]
+//! #   unsafe { signal_hook::__emulate_kill(libc::getpid(), signal_hook::SIGUSR1) };
+//! #   #[cfg(not(windows))]
 //! #   unsafe { libc::kill(libc::getpid(), signal_hook::SIGUSR1) };
 //!     'outer: loop {
 //!         // Pick up signals that arrived since last time
@@ -52,15 +55,12 @@
 use std::borrow::Borrow;
 use std::io::Error;
 use std::iter::Enumerate;
-use std::os::unix::io::AsRawFd;
-use std::os::unix::net::UnixStream;
 use std::slice::Iter;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
-use libc::{self, c_int};
+use libc::c_int;
 
-use pipe;
 use SigId;
 
 /// Maximal signal number we support.
@@ -70,14 +70,13 @@ const MAX_SIGNUM: usize = 128;
 struct Waker {
     pending: Vec<AtomicBool>,
     closed: AtomicBool,
-    read: UnixStream,
-    write: UnixStream,
+    inner: self::waker_impl::Waker,
 }
 
 impl Waker {
     /// Sends a wakeup signal to the internal wakeup pipe.
     fn wake(&self) {
-        pipe::wake(self.write.as_raw_fd());
+        self.inner.wake();
     }
 }
 
@@ -163,13 +162,12 @@ impl Signals {
         I: IntoIterator<Item = S>,
         S: Borrow<c_int>,
     {
-        let (read, write) = UnixStream::pair()?;
+        let inner = waker_impl::Waker::new()?;
         let pending = (0..MAX_SIGNUM).map(|_| AtomicBool::new(false)).collect();
         let waker = Arc::new(Waker {
             pending,
             closed: AtomicBool::new(false),
-            read,
-            write,
+            inner,
         });
         let ids = (0..MAX_SIGNUM).map(|_| None).collect();
         let me = Self {
@@ -231,19 +229,7 @@ impl Signals {
         if self.waker.closed.load(Ordering::SeqCst) {
             return false;
         }
-        const SIZE: usize = 1024;
-        let mut buff = [0u8; SIZE];
-        let res = unsafe {
-            // We ignore all errors on purpose. This should not be something like closed file
-            // descriptor. It could EAGAIN, but that's OK in case we say MSG_DONTWAIT. If it's
-            // EINTR, then it's OK too, it'll only create a spurious wakeup.
-            libc::recv(
-                self.waker.read.as_raw_fd(),
-                buff.as_mut_ptr() as *mut libc::c_void,
-                SIZE,
-                if wait { 0 } else { libc::MSG_DONTWAIT },
-            )
-        };
+        let res = self.waker.inner.flush(wait);
 
         if self.waker.closed.load(Ordering::SeqCst) {
             // Wake any other sleeping ends
@@ -251,7 +237,7 @@ impl Signals {
             // anyway).
             self.waker.wake();
         }
-        res > 0
+        res
     }
 
     /// Returns an iterator of already received signals.
@@ -431,13 +417,145 @@ impl<'a> Iterator for Forever<'a> {
     }
 }
 
+#[cfg(not(windows))]
+mod waker_impl {
+    use libc;
+    use std::io::Error;
+    use std::os::unix::io::AsRawFd;
+    use std::os::unix::io::RawFd;
+    use std::os::unix::net::UnixStream;
+
+    #[cfg(feature = "mio-support")]
+    use mio::unix::EventedFd;
+
+    use pipe;
+
+    #[derive(Debug)]
+    pub(super) struct Waker {
+        read: UnixStream,
+        write: UnixStream,
+        read_fd: RawFd,
+    }
+
+    impl Waker {
+        pub(super) fn new() -> Result<Self, Error> {
+            let (read, write) = UnixStream::pair()?;
+            let read_fd = read.as_raw_fd();
+            Ok(Self {
+                read,
+                write,
+                read_fd,
+            })
+        }
+
+        /// Sends a wakeup signal to the internal wakeup pipe.
+        pub(super) fn wake(&self) {
+            pipe::wake(self.write.as_raw_fd());
+        }
+
+        /// Reads data from the internal self-pipe.
+        ///
+        /// If `wait` is `true` and there are no data in the self pipe, it blocks until some come.
+        ///
+        /// Returns weather it successfully read something.
+        pub(super) fn flush(&self, wait: bool) -> bool {
+            const SIZE: usize = 1024;
+            let mut buff = [0u8; SIZE];
+            let res = unsafe {
+                // We ignore all errors on purpose. This should not be something like closed file
+                // descriptor. It could EAGAIN, but that's OK in case we say MSG_DONTWAIT. If it's
+                // EINTR, then it's OK too, it'll only create a spurious wakeup.
+                libc::recv(
+                    self.read.as_raw_fd(),
+                    buff.as_mut_ptr() as *mut libc::c_void,
+                    SIZE,
+                    if wait { 0 } else { libc::MSG_DONTWAIT },
+                )
+            };
+
+            res > 0
+        }
+
+        #[cfg(feature = "mio-support")]
+        pub(super) fn evented(&self) -> EventedFd {
+            EventedFd(&self.read_fd)
+        }
+    }
+}
+
+#[cfg(windows)]
+mod waker_impl {
+    // Unlike `sigaction`, `SetConsoleCtrlHandler` launches a separate thread for handlers.
+    // Therefore we don't need to be AS-safe and can just use an ordinary in-process synchronization primitives.
+
+    use std::io::Error;
+    use std::mem;
+    use std::sync::{Condvar, Mutex};
+
+    #[cfg(feature = "mio-support")]
+    use mio::{Ready, Registration, SetReadiness};
+
+    #[derive(Debug)]
+    pub(super) struct Waker {
+        waken: Mutex<bool>,
+        cond: Condvar,
+        #[cfg(feature = "mio-support")]
+        registration: Registration,
+        #[cfg(feature = "mio-support")]
+        set_readiness: SetReadiness,
+    }
+
+    impl Waker {
+        pub(super) fn new() -> Result<Self, Error> {
+            #[cfg(feature = "mio-support")]
+            let (registration, set_readiness) = Registration::new2();
+            Ok(Waker {
+                waken: Mutex::new(false),
+                cond: Condvar::new(),
+                #[cfg(feature = "mio-support")]
+                registration,
+                #[cfg(feature = "mio-support")]
+                set_readiness,
+            })
+        }
+
+        /// Sends a wakeup signal to the internal wakeup pipe.
+        pub(super) fn wake(&self) {
+            let mut waken = self.waken.lock().unwrap();
+            if *waken == false {
+                *waken = true;
+                self.cond.notify_one();
+            }
+            // How do we handle error in set_readiness?
+            #[cfg(feature = "mio-support")]
+            self.set_readiness.set_readiness(Ready::readable()).ok();
+        }
+
+        /// Reads data from the internal self-pipe.
+        ///
+        /// If `wait` is `true` and there are no data in the self pipe, it blocks until some come.
+        ///
+        /// Returns weather it successfully read something.
+        pub(super) fn flush(&self, wait: bool) -> bool {
+            let mut waken = self.waken.lock().unwrap();
+            while wait && !*waken {
+                waken = self.cond.wait(waken).unwrap();
+            }
+            mem::replace(&mut *waken, false)
+        }
+
+        #[cfg(feature = "mio-support")]
+        pub(super) fn evented(&self) -> &Registration {
+            &self.registration
+        }
+    }
+}
+
 #[cfg(feature = "mio-support")]
 mod mio_support {
     use std::io::Error;
-    use std::os::unix::io::AsRawFd;
 
     use mio::event::Evented;
-    use mio::unix::EventedFd;
     use mio::{Poll, PollOpt, Ready, Token};
 
     use super::Signals;
@@ -450,7 +568,10 @@ mod mio_support {
             interest: Ready,
             opts: PollOpt,
         ) -> Result<(), Error> {
-            EventedFd(&self.waker.read.as_raw_fd()).register(poll, token, interest, opts)
+            self.waker
+                .inner
+                .evented()
+                .register(poll, token, interest, opts)
         }
 
         fn reregister(
@@ -460,11 +581,16 @@ mod mio_support {
             interest: Ready,
             opts: PollOpt,
         ) -> Result<(), Error> {
-            EventedFd(&self.waker.read.as_raw_fd()).reregister(poll, token, interest, opts)
+            self.waker
+                .inner
+                .evented()
+                .reregister(poll, token, interest, opts)
         }
 
         fn deregister(&self, poll: &Poll) -> Result<(), Error> {
-            EventedFd(&self.waker.read.as_raw_fd()).deregister(poll)
+            // The inherent method is deprecated but the trait method isn't
+            #[cfg_attr(windows, allow(deprecated))]
+            self.waker.inner.evented().deregister(poll)
         }
     }
 
@@ -477,6 +603,11 @@ mod mio_support {
 
         use super::*;
 
+        #[cfg(windows)]
+        use __emulate_kill as kill;
+        #[cfg(not(windows))]
+        use libc::kill;
+
         #[test]
         fn mio_wakeup() {
             let signals = Signals::new(&[::SIGUSR1]).unwrap();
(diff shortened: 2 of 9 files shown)


vorner:
Hello.

Thanks for the PR. I'll get to reviewing in few days, unfortunately I don't have the time right now.

However, I wonder if there are such differences in API and such, if it makes sense to have it in the same crate at all. The alternative would be to have a separate crate with the same or similar API and have the selection at the caller side. If we provided an example how to do that, it could work well too.

I don't have a preference (definitely not before reading it), but can I ask what advantages or disadvantages you see in including it here?

src/pipe_windows.rs
```diff
@@ -0,0 +1,204 @@
...
+//! # fn do_something_that_may_be_interrupted_by_ctrl_c() {}
+//!     do_something_that_may_be_interrupted_by_ctrl_c();
+//!     let mut buff = [0];
+//! # if false {
+//!     read.read_exact(&mut buff)?;
+//! # }
+//!     println!("Happily terminating");
+//!     Ok(())
+//! }
+//! ```
```
vorner: This seems the documentation here talks mostly about unix too ‒ there are things like `UnixStream` mentioned.

src/pipe_windows.rs
```diff
@@ -0,0 +1,204 @@
...
+    let pipe = SendSocket(pipe);
+    unsafe { ::register(signal, move || wake_socket(pipe.0)) }
+}
+
+/// Registers a write to a self-pipe whenever there's the signal.
+///
+/// The ownership of pipe is taken and will be closed whenever the created action is unregistered.
+///
+/// Note that if you want to register the same pipe for multiple signals, there's `try_clone`
+/// method on many unix socket primitives.
```
vorner: Again, speaking about unix sockets.

src/pipe_windows.rs
```diff
@@ -0,0 +1,204 @@
...
+pub fn register_socket<P>(signal: c_int, pipe: P) -> Result<SigId, Error>
+where
+    P: AsRawSocket + Send + Sync + 'static,
+{
+    let action = move || wake_socket(pipe.as_raw_socket());
+    unsafe { ::register(signal, action) }
+}
```
vorner: It seems a bit unfortunate that there are two sets of register functions here, one for handles, another for sockets. If you adopt the above suggestion about abstracting the differences away by a custom trait, could the AsPipe trait be implemented for both sets at once?

src/pipe_windows.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn pipe_notify() -> Result<(), ::std::io::Error> {
+        let (mut read, write) = {
+            let mut read: HANDLE = ptr::null_mut();
+            let mut write: HANDLE = ptr::null_mut();
+            let success = unsafe { CreatePipe(&mut read, &mut write, ptr::null_mut(), 1024) } != 0;
+            assert!(success);
+            let read = unsafe { File::from_raw_handle(read as RawHandle) };
+            let write = unsafe { File::from_raw_handle(write as RawHandle) };
+            (read, write)
+        };
```
vorner: These blocks are usually introduced to appease the borrow checker. But I don't see why it is needed here. Can you point me to what needs to go out of scope here?

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+// SIGFPE = 8
+// SIGSEGV = 11
+// SIGTERM = 15
+// SIGBREAK = 21
+// SIGABRT = 22
+pub use libc::{SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV, SIGTERM};
+
+const DUMMY_OFFSET: c_int = 64;
+
+// Define as Linux x86 signal number + 10000
```
vorner: The comment is not true.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+/// Dummy value for SIGSYS which is missing in Windows.
+pub const SIGSYS: c_int = DUMMY_OFFSET + 31;
+/// Dummy value for SIGTRAP which is missing in Windows.
+pub const SIGTRAP: c_int = DUMMY_OFFSET + 5;
+/// Dummy value for SIGUSR1 which is missing in Windows.
+pub const SIGUSR1: c_int = DUMMY_OFFSET + 10;
+/// Dummy value for SIGUSR2 which is missing in Windows.
+pub const SIGUSR2: c_int = DUMMY_OFFSET + 12;
+/// Dummy value for SIGWINCH which is missing in Windows.
+pub const SIGWINCH: c_int = DUMMY_OFFSET + 28;
```
vorner: Can you explain why these constants are interesting and should exist? If I understand it correctly, they are not available on Windows. Therefore, any use of these will likely lead to an error, or at least nothing will happen.

If it's the case, I'd prefer not to have them at all, so the programmer is aware at compile time that functionality is not present, instead of having the program terminate with an error or not work.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+
+/// List of forbidden signals.
+///
+/// Some signals are impossible to replace according to POSIX and some are so special that this
+/// library refuses to handle them (eg. SIGSEGV). The routines panic in case registering one of
+/// these signals is attempted.
+///
+/// See [`register`](fn.register.html).
```
vorner: Speaking about posix in windows implementation? 😇

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+impl GlobalDataState {
+    fn new() -> Self {
+        Self {
+            action_id: 0,
+            all_signals: HashMap::new(),
+        }
+    }
+}
```
vorner: This gymnastics is needed on unix systems to make sure only async-signal-safe primitives are used in the signal handlers. I believe it is fine to use something like `std::sync::Once` or `lazy_static` here and get rid of the whole lot of code.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+            // returning TRUE here has two effects:
+            // 1. Default behavior is prevented.
+            // 2. Handlers registered earlier than this are not invoked.
+            //
+            // Ideally we want to prevent default behavior while keeping
+            // old callbacks, but it's impossible.
+            //
+            // And we may want to return FALSE if `all_signals` is empty.
+            // I didn't do so to align with the unix implementation.
+            return TRUE;
```
vorner: What is the default behaviour? Because, this is actually not aligned with the unix implementation where the default is prevented, but previous signal handlers are called.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+
+fn to_signal(ctrl_type: DWORD) -> Option<c_int> {
+    if ctrl_type == CTRL_C_EVENT {
+        Some(SIGINT)
+    } else if ctrl_type == CTRL_BREAK_EVENT {
+        Some(SIGTERM)
+    } else {
+        None
+    }
+}
```
vorner: So, effectively, there are only *two* signals on windows/this implementation? The rest will just not do anything at all? Then it is kind of confusing that they are exported.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+
+/// Registers an arbitrary action for the given signal.
+///
+/// This makes sure there's a signal handler for the given signal. It then adds the action to the
+/// ones called each time the signal is delivered. If multiple actions are set for the same signal,
+/// all are called, in the order of registration.
+///
+/// If there was a previous signal handler for the given signal, it is chained ‒ it will be called
+/// as part of this library's signal handler, before any actions set through this function.
```
vorner: As mentioned above, it (if I understand it correctly) is not chained.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+/// descriptors. Calling program's own functions consisting only of these is OK, as is manipulating
+/// program's variables ‒ however, as the action can be called on any thread that does not have the
+/// given signal masked (by default no signal is masked on any thread), and mutexes are a no-go,
+/// this is harder than it looks like at first.
+///
+/// As panicking from within a signal handler would be a panic across FFI boundary (which is
+/// undefined behavior), the passed handler must not panic.
+///
+/// If you find these limitations hard to satisfy, choose from the helper functions in submodules
+/// of this library ‒ these provide safe interface to use some common signal handling patters.
```
vorner: Again, this probably doesn't apply.

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+    };
+
+    let all_signals = state
+        .all_signals
+        .entry(signal)
+        .or_insert_with(BTreeMap::new);
+    all_signals.insert(action_id, Box::new(action));
+
+    Ok(sig_id)
+}
```
vorner: Reading this function, I have the impression there must be something missing. Who registers the `handler` function into the OS?

src/windows.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    let data = GlobalData::ensure();
+    let state = data.state.lock().unwrap();
+    if let Some(all_signals) = state.all_signals.get(&signal) {
+        for action in all_signals.values() {
+            action();
+        }
+    }
+    0
+}
```
vorner: I understand you want to use this inside the tests. But AFAIK this actually doesn't test anything (or, only part), because it doesn't go through the signal/console control.

Is there no way to send the real thing in Windows?

qnighy:
Thanks for reviewing! I'll take a look at them later. I'm still confused about relationship between Win32API `SetConsoleCtrlHandler`/`GenerateConsoleCtrlEvent` and CRT `signal`/`raise`, so I'll further investigate it. My latest understanding is that:

- [`signal`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=vs-2019) is ISO C compatible, so it can be used as a lesser alternative to `sigaction`. However, [Windows `signal` calls the registered handler only once](https://wiki.sei.cmu.edu/confluence/display/c/SIG01-C.+Understand+implementation-specific+details+regarding+signal+handler+persistence) except SIGFPE.
- There's a corresponding [`raise`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/raise?view=vs-2019) function, but `kill` isn't.
- [`SetConsoleCtrlHandler`](https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler) and `signal` overlaps for Ctrl+C(SIGINT) and Ctrl+Break(SIGBREAK). Ctrl+Break wasn't SIGTERM despite my first impression. `signal` forwards handler registrations for these signals to `SetConsoleCtrlHandler`. Other signals are `signal`-specific and other console controls are `SetConsoleCtrlHandler`-specific.
- There's a corresponding [`GenerateConsoleCtrlEvent`](https://docs.microsoft.com/en-us/windows/console/generateconsolectrlevent), which I couldn't find at first glance. This seems to be a `kill`-equivalent for `SetConsoleCtrlHandler`.

vorner:
These are interesting finds.

May I have a suggestion? It would be nice to  have a support for all the existing signals there. The `signal` call is really similar to `sigaction` (the latter one is simply an extension and clarification). I'd propose to go in stages and see how it works:

* Modify the signal-hook-registry to support windows by local `#[cfg(…)]` to use `signal` there and re-register the signal as the first thing in the handler (which still has the race condition, but it's the best thing available right now). For this to be able to compile & test, I'd propose disabling all the problematic parts of signal-hook on windows and postpone it for further PRs.
* Add support for the ConsoleCtrlHandler, because in case of the SIGINT/SIGBREAK, it is probably better ‒ so have special casing for these. But it still could be local cfg-modifications in the main code, reusing most of them.
* Add modifications to the pipe & iterator modules.

Mostly, what I propose is to bring your changes in, but by smaller parts so it's easier to discuss them and iterate on them. Would that strategy work for you?

signal-hook-registry/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 )]
 #![deny(missing_docs, warnings)]
 #![allow(unknown_lints, bare_trait_objects)]
+#![cfg(not(windows))]
```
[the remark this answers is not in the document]
qnighy: [`winapi`](https://github.com/retep998/winapi-rs/blob/0.3.7/src/lib.rs#L6) and [`nix`](https://github.com/nix-rust/nix/blob/v0.14.1/src/lib.rs#L6) do this. I guess it's beneficial to maximally avoid `cfg`s in `Cargo.toml`.

src/iterator.rs
```diff
@@ -431,13 +417,154 @@ impl<'a> Iterator for Forever<'a> {
...
+}
+
+#[cfg(windows)]
+mod waker_impl {
+    // Unlike `sigaction`, `SetConsoleCtrlHandler` launches a separate thread for handlers.
+    // Therefore we don't need to be AS-safe and can just use an ordinary in-process synchronization primitives.
+
+    use std::io::Error;
+    use std::mem;
+    use std::sync::{Condvar, Mutex};
```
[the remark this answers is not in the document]
qnighy: I thought `Mutex` with `Condvar` is "enough" because it's more primitive. But I see that `mpsc` might be more suitable here because it's parallel with `#[cfg(unix)]` implementaion.

src/iterator.rs
```diff
@@ -431,13 +417,154 @@ impl<'a> Iterator for Forever<'a> {
...
+    #[cfg(feature = "mio-support")]
+    use mio::{Ready, Registration, SetReadiness};
+
+    use self::WakeState::*;
+
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    enum WakeState {
+        Waken,
+        NotWaken,
+    }
```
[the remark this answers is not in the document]
qnighy: :see_no_evil:

src/iterator.rs
```diff
@@ -431,13 +417,154 @@ impl<'a> Iterator for Forever<'a> {
...
+                waken: Mutex::new(NotWaken),
+                cond: Condvar::new(),
+                #[cfg(feature = "mio-support")]
+                registration,
+                #[cfg(feature = "mio-support")]
+                set_readiness,
+            })
+        }
+
+        /// Sends a wakeup signal to the internal wakeup pipe.
```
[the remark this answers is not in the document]
qnighy: No, the comments here were copies from the original implementation.

src/lib.rs
```diff
@@ -128,15 +128,35 @@ extern crate mio;
...
 extern crate tokio_reactor;
+#[cfg(windows)]
+extern crate winapi;
 
 pub mod flag;
 pub mod iterator;
+#[cfg(not(windows))]
 pub mod pipe;
+#[cfg(windows)]
+pub mod pipe_windows;
```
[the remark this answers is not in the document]
qnighy: I'm not sure we can do this as there must be a reason they made `AsRawFd` OS-specific, but I want to give it a try too.

src/pipe_windows.rs
qnighy: I guess not that easy due to coherence.

src/pipe_windows.rs
qnighy: Block here was for aligning with this line:

```rust
let (read, write) = UnixStream::pair();
```

src/windows.rs
qnighy: 🙈

src/windows.rs
qnighy: I tended to this to have "a unified interface" here to minimize `#[cfg]` in implementation and documentation. I started to think that this isn't a good idea, but still want to be able to use `SIGUSR1` in tests and doc-tests.

src/windows.rs
qnighy: Ah, I misunderstood the reason why `lazy_static` isn't used here.

src/windows.rs
qnighy: https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler

> When a console process receives any of the control signals, its handler functions are called on a last-registered, first-called basis until one of the handlers returns TRUE. If none of the handlers returns TRUE, the default handler is called.

Therefore, TRUE prevents both previous handlers and default handlers. FALSE calls both previous handlers and default handlers (unless one of these "previous handlers" returns TRUE)

src/windows.rs
qnighy: Unlike `signal`/`sigaction`, `SetConsoleCtrlHandler` registry is not signal-specific. Therefore I registered the handler at `GlobalData` initialization.

src/windows.rs
qnighy: As mentioned, I've missed it and it actually exists.

qnighy:
> `register_unchecked` can be implemented, `register_sigaction` probably can't

`register_unchecked` is able to receive `siginfo_t`, which doesn't exist in Windows. It might be good to have something like `register_unchecked_without_sigaction`.

qnighy:
> May I have a suggestion?

I like these suggestions. I'll give them a try in another branch.

vorner:
Hello

I just want to make sure this is not deadlocked (you waiting on me to do something and I waiting on you 😇). What are the plans for further progress? Will you prepare another branch with another part (maybe enabling the signal iterator or the pipes), or should we try to update this branch?

Or, would you like some help (not with Windows for sure, but with some other things)?

vorner:
@qnighy Unless I hear from you soon, I'll probably take what's in this branch and try to polish it and get it in somehow.

hskang9:
@vorner  is this on progress? I want to work on this.

qnighy:
@vorner @hskang9 sorry for the long silence! Currently it's stale on my side. I appreciate someone trying to push this forward or having another attempt, but I have one thing to note: #19 has a different approach than this PR (#18) so assumptions are also different. We can ignore AS-safety when using `SetConsoleCtrlHandler` (because the handlers are run in a different thread) but `signal` doesn't have such a guarantee, as far as I know. Therefore, iterator handling would be different from the status quo of #18.

vorner:
I didn't get around to it and I'll be happy if either of you pushes this over the finish line.

As noted in the discussions and in #19, it'll be great if this gets reviewed an in in smaller chunks and if the API is as close as possible on Windows too, so users can just use it and not put a lot of `#[cfg(...)]` directives in their code.

vorner:
Hello? I'm a bit confused here what the status is, it seems to be somehow stuck.

Is anyone of you actually working or wants to work on this? @qnighy @hskang9? Is there something I can help with?

hskang9:
Hey @vorner, I just finished Hacktoberfest, and I am getting back to this. I was lost after finding out the syscall of Windows and Unix/Linux was different. so my questions are:

- what is the code which loads up all syscall? 

- If there is the code which has config for different operating systems, where can I find it?

vorner:
I'm not really sure what you're asking about. Are you asking about signals or syscalls? What do you mean with „loads up“?

Ability to register actions to signals is already part of the library even for windows. The code is in the `signal-hook-registry` subcrate and re-exported from `signal-hook` itself. The list of signals that are supported is in `src/lib.rs`.

What is still missing is making the `pipe` and `iterator` modules work on windows too, because these use platform specific primitives. I believe in case of the `iterator` module it'll be possible to get the exact same interface while hiding something slightly else inside (maybe just replacing the unix socket pair with TCP socket pair on windows). The `pipe` module might be a bit more tricky, because the interface won't be 100% the same ‒ there's no `AsRawFd` on Windows. Nevertheless, it would be great if the function calls would be named the same way and if they could accept the same parameters (eg. pipes, sockets, ...), so user code doesn't have to be changed. There's no `AsRawFd` on Windows, but there's something like `AsRawHandle` and that might work.

hskang9:
yeah the current problem of the `signal_hook` is that the library tries to use `AsRawFd` to operate in Windows. If there is a compile option for target os in there, that would fix most of the problem I think.

vorner:
Well, you can use the `#[cfg(windows)]` and `#[cfg(not(windows))]` as necessary, both in implementation and the API. Something like this:

https://github.com/vorner/signal-hook/blob/master/src/lib.rs#L156

hskang9:
Yes, so far libc gets all the necessary commands for its target OS. However, the [`iterator.rs`](https://github.com/vorner/signal-hook/blob/af449e8a1fb47477bf6c8d5424ed094fab28727a/src/iterator.rs#L55) imports unix standard library which does not exist on windows and creates error. I will add compile option tags there and let's see that would fix most of the problem.

vorner:
Any success yet?

vorner:
Hello. Any progress?

vorner:
I'll close this because of prolonged inactivity. I'd still be happy to have better Windows support in there, but this one doesn't seem to be moving :-(

PR #19, opened 2019-07-21
Implement register and flag::register for Windows.

qnighy:
This is another attempt at Windows support in #18.

Main differences from #18:

- `pipe`/`iterator` supports are postponed.
- Not using `SetConsoleCtrlHandler` directly. Instead, use `signal`/`raise` directly.
- We no longer have the `__emulate_kill` workaround. `raise` works.

As for #18, confirmed to catch Ctrl-C in some cases, but not all cases.

### DIFF
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -228,8 +228,17 @@ extern "C" fn handler(sig: c_int) {
         // Problems:
         // - It's racy. But this is inevitably racy in Windows.
         // - Interacts poorly with handlers outside signal-hook-registry.
-        unsafe {
-            libc::signal(sig, handler as sighandler_t);
+        let old = unsafe { libc::signal(sig, handler as sighandler_t) };
+        if old == SIG_ERR {
+            // MSDN doesn't describe which errors might occur,
+            // but we can tell from the Linux manpage that
+            // EINVAL (invalid signal number) is mostly the only case.
+            // Therefore, this branch must not occur.
+            // In any case we can do nothing useful in the signal handler,
+            // so we're going to abort silently.
+            unsafe {
+                libc::abort();
+            }
         }
     }
 
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -649,7 +649,7 @@ mod tests {
         unsafe {
             pid = libc::getpid();
             register_sigaction(SIGUSR2, action).unwrap();
-            libc::kill(pid, SIGUSR2);
+            libc::raise(SIGUSR2);
         }
         for _ in 0..10 {
             thread::sleep(Duration::from_millis(100));
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -341,12 +341,12 @@ fn without_signal<F: FnOnce() -> Result<(), Error>>(signal: c_int, f: F) -> Resu
 /// these signals is attempted.
 ///
 /// See [`register`](fn.register.html).
-pub const FORBIDDEN: &[c_int] = _FORBIDDEN;
+pub const FORBIDDEN: &[c_int] = FORBIDDEN_IMPL;
 
 #[cfg(windows)]
-const _FORBIDDEN: &[c_int] = &[SIGILL, SIGFPE, SIGSEGV];
+const FORBIDDEN_IMPL: &[c_int] = &[SIGILL, SIGFPE, SIGSEGV];
 #[cfg(not(windows))]
-const _FORBIDDEN: &[c_int] = &[SIGKILL, SIGSTOP, SIGILL, SIGFPE, SIGSEGV];
+const FORBIDDEN_IMPL: &[c_int] = &[SIGKILL, SIGSTOP, SIGILL, SIGFPE, SIGSEGV];
 
 /// Registers an arbitrary action for the given signal.
 ///
@@ -439,7 +439,7 @@ pub unsafe fn register<F>(signal: c_int, action: F) -> Result<SigId, Error>
 where
     F: Fn() + Sync + Send + 'static,
 {
-    register_sigaction_(signal, move |_: &_| action())
+    register_sigaction_impl(signal, move |_: &_| action())
 }
 
 /// Register a signal action.
@@ -452,10 +452,10 @@ pub unsafe fn register_sigaction<F>(signal: c_int, action: F) -> Result<SigId, E
 where
     F: Fn(&siginfo_t) + Sync + Send + 'static,
 {
-    register_sigaction_(signal, action)
+    register_sigaction_impl(signal, action)
 }
 
-unsafe fn register_sigaction_<F>(signal: c_int, action: F) -> Result<SigId, Error>
+unsafe fn register_sigaction_impl<F>(signal: c_int, action: F) -> Result<SigId, Error>
 where
     F: Fn(&siginfo_t) + Sync + Send + 'static,
 {
@@ -464,7 +464,7 @@ where
         "Attempted to register forbidden signal {}",
         signal,
     );
-    register_unchecked_(signal, action)
+    register_unchecked_impl(signal, action)
 }
 
 /// Register a signal action without checking for forbidden signals.
@@ -475,7 +475,7 @@ pub unsafe fn register_signal_unchecked<F>(signal: c_int, action: F) -> Result<S
 where
     F: Fn() + Sync + Send + 'static,
 {
-    register_unchecked_(signal, move |_: &_| action())
+    register_unchecked_impl(signal, move |_: &_| action())
 }
 
 /// Register a signal action without checking for forbidden signals.
@@ -492,10 +492,10 @@ pub unsafe fn register_unchecked<F>(signal: c_int, action: F) -> Result<SigId, E
 where
     F: Fn(&siginfo_t) + Sync + Send + 'static,
 {
-    register_unchecked_(signal, action)
+    register_unchecked_impl(signal, action)
 }
 
-unsafe fn register_unchecked_<F>(signal: c_int, action: F) -> Result<SigId, Error>
+unsafe fn register_unchecked_impl<F>(signal: c_int, action: F) -> Result<SigId, Error>
 where
     F: Fn(&siginfo_t) + Sync + Send + 'static,
 {
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -42,6 +42,15 @@
 //! require dependencies that don't build there, so tests need newer Rust version (they are run on
 //! stable).
 //!
+//! # Portability
+//!
+//! This crate includes a limited support for Windows, based on `signal`/`raise` in the CRT.
+//! There are differences in both API and behavior:
+//!
+//! - Due to lack of `siginfo_t`, we don't provide `register_sigaction` or `register_unchecked`.
+//! - Due to lack of signal blocking, there's a race condition.
+//!   After the call to `signal`, there's a moment where we miss a signal.
+//!
 //! [signal-hook]: https://docs.rs/signal-hook
 //! [async-signal-safe]: http://www.man7.org/linux/man-pages/man7/signal-safety.7.html
 
@@ -52,6 +61,7 @@ use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap};
 use std::io::Error;
 use std::mem;
+#[cfg(not(windows))]
 use std::ptr;
 // Once::new is now a const-fn. But it is not stable in all the rustc versions we want to support
 // yet.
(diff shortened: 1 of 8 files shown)


signal-hook-registry/src/lib.rs
```diff
@@ -42,6 +42,15 @@
...
 //! stable).
 //!
+//! # Portability
+//!
+//! This crate includes a limited support for Windows, based on `signal`/`raise` in the CRT.
+//! There are differences in both API and behavior:
+//!
+//! - Due to lack of `siginfo_t`, we don't provide `register_sigaction` or `register_unchecked`.
+//! - Due to lack of signal blocking, there's a race condition.
+//!   After the call to `signal`, there's a moment where we miss a signal.
```
vorner: I'd suggest listing the consequences here ‒ „If this happens, the default handler is called instead“.

signal-hook-registry/src/lib.rs
```diff
@@ -167,6 +213,38 @@ impl GlobalData {
...
+    }
+
+    let signals = GlobalData::get().all_signals.peek_signal_safe();
+
+    if let Some(ref slot) = signals.get(&sig) {
+        let fptr = slot.prev;
+        if fptr != 0 && fptr != SIG_DFL && fptr != SIG_IGN {
+            // FFI ‒ calling the original signal handler.
+            unsafe {
+                let action = mem::transmute::<usize, extern "C" fn(c_int)>(fptr);
```
vorner: I wonder, if it is possible to convert in one direction with `as` (eg. the `handler as sighandler_t`), would it work the other way around too, instead of using transmute?

signal-hook-registry/src/lib.rs
```diff
@@ -247,7 +334,12 @@ fn without_signal<F: FnOnce() -> Result<(), Error>>(signal: c_int, f: F) -> Resu
 /// these signals is attempted.
 ///
 /// See [`register`](fn.register.html).
-pub const FORBIDDEN: &[c_int] = &[SIGKILL, SIGSTOP, SIGILL, SIGFPE, SIGSEGV];
+pub const FORBIDDEN: &[c_int] = _FORBIDDEN;
```
vorner: Why is there the indirection? To reuse the doc string?

I'd probably prefer to use something like `FORBIDDEN_IMPL` or `FORBIDDEN_INNER` instead of surrounding names with underscores.

signal-hook-registry/src/lib.rs
```diff
@@ -167,6 +213,38 @@ impl GlobalData {
...
+#[cfg(windows)]
+extern "C" fn handler(sig: c_int) {
+    if sig != SIGFPE {
+        // Windows CRT `signal` resets handler every time, unless for SIGFPE.
+        // Reregister the handler to retain maximal compatibility.
+        // Problems:
+        // - It's racy. But this is inevitably racy in Windows.
+        // - Interacts poorly with handlers outside signal-hook-registry.
+        unsafe {
+            libc::signal(sig, handler as sighandler_t);
```
vorner: I guess there should be some error handling here.

I'm not 100% sure what to do if there's an error, though. Maybe doing `libc::abort()` and commenting that there's nothing better to be done in such case and that it should probably not happen anyway (because we already *had* the signal), but I'd prefer not to leave the application in somewhat half-broken state.

signal-hook-registry/src/lib.rs
```diff
@@ -369,7 +480,15 @@ where
...
 /// function still allows you to do it.
+#[cfg(not(windows))]
 pub unsafe fn register_unchecked<F>(signal: c_int, action: F) -> Result<SigId, Error>
+where
+    F: Fn(&siginfo_t) + Sync + Send + 'static,
+{
+    register_unchecked_(signal, action)
+}
+
+unsafe fn register_unchecked_<F>(signal: c_int, action: F) -> Result<SigId, Error>
```
vorner: Again, I'd propose to name it something like `register_unchecked_impl` or such, not just give it a tail-underscore.

signal-hook-registry/src/lib.rs
```diff
@@ -435,23 +554,57 @@ mod tests {
     use std::thread;
     use std::time::Duration;
 
+    #[cfg(not(windows))]
     use libc::{pid_t, SIGUSR1, SIGUSR2};
 
+    #[cfg(windows)]
+    use libc::SIGTERM as SIGUSR1;
+    #[cfg(windows)]
+    use libc::SIGTERM as SIGUSR2;
```
vorner: :thinking: I wonder why the tests used separate signals. Could something bad happen? One test sending a signal to another? Invoke the race condition? Because tests are allowed to run in parallel.

I don't have a solution ready and this is inside tests only, so I'd propose to simply add a FIXME note warning about it; if the tests fail because of the race condition, such note will help understand what happened.

src/flag.rs
```diff
@@ -82,10 +85,17 @@
...
+//! #   #[cfg(windows)]
+//! #   signal_hook::flag::register(signal_hook::SIGTERM, Arc::clone(&got))?;
 //!     unsafe {
+//! #       #[cfg(not(windows))]
 //!         let pid = libc::getpid();
+//! #       #[cfg(not(windows))]
 //!         libc::kill(pid, signal_hook::SIGUSR1);
+//! #       #[cfg(windows)]
+//! #       libc::raise(signal_hook::SIGTERM);
 //!     }
```
vorner: Let's simplify this one :-). Let's use SIGTERM and raise in both windows and linux versions.

src/flag.rs
```diff
@@ -187,7 +203,10 @@ mod tests {
     fn register_unregister() {
         // When we register the action, it is active.
         let flag = Arc::new(AtomicBool::new(false));
+        #[cfg(not(windows))]
         let signal = register(::SIGUSR1, Arc::clone(&flag)).unwrap();
+        #[cfg(windows)]
+        let signal = register(::SIGTERM, Arc::clone(&flag)).unwrap();
```
vorner: I believe these tests were using USR1 simply to leave SIGTERM alone, so I could CTRL+C the tests if the got into an infinite loop or something. Killing them with SIGKILL works too and is probably better than having these windows/non-windows conditions, so let's use SIGTERM in both.

src/flag.rs
```diff
@@ -164,8 +176,12 @@ mod tests {
 
     fn self_signal() {
         unsafe {
+            #[cfg(not(windows))]
             let pid = libc::getpid();
+            #[cfg(not(windows))]
             libc::kill(pid, ::SIGUSR1);
+            #[cfg(windows)]
+            libc::raise(::SIGTERM);
```
vorner: Here I actually prefer the way you wrote it (with the windows/not-windows conditions), because the getpid/kill combination allows delivering the signal to arbitrary thread, not just the current one. This allows some more bugs to manifest.

But a comment giving the rationale would be nice (eg. explaining the advantage of getpid/kill, vs Windows not having kill)

src/lib.rs
```diff
@@ -85,8 +85,20 @@
...
 //!
-//! Windows has some limited support for signals, patches to include support in this library are
-//! welcome.
+//! This crate includes a limited support for Windows, based on `signal`/`raise` in the CRT.
+//! There are differences in both API and behavior:
+//!
+//! - `iterator` and `pipe` are not yet implemented.
+//! - We have only a few signals: `SIGABRT`, `SIGFPE`, `SIGILL`, `SIGINT`, `SIGSEGV` and `SIGTERM`.
+//!   - CRT headers also define `SIGABRT_COMPAT` and `SIGBREAK`,
+//!     but it's not exported from libc yet.
```
vorner: :thinking: Would you know if the constants are guaranteed to be the same everywhere? We could define them ourselves in such case. Not necessarily in scope of this PR, though.

tests/iterator.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(not(windows))]
+
```
vorner: AFAIK this is not needed any more, the condition is already on the `pub mod` in `lib.rs`.

tests/tokio.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(not(windows))]
+
```
vorner: Similar.

signal-hook-registry/src/lib.rs
qnighy: That must not be the case, because `usize` allows 0 while `fn` doesn't allow 0 (instant UB, IIRC)

signal-hook-registry/src/lib.rs
qnighy: You mean to use `SIGUSR1` in both cases (in the tests for POSIX)?

src/flag.rs
qnighy: Isn't it good to choose a "user-defined" signal? I chose `SIGTERM` here because `SIGTERM` is said not to be generated under Windows, and unfortunately there's no `SIGUSR*` in Windows. https://docs.microsoft.com/ja-jp/cpp/c-runtime-library/reference/signal?view=vs-2019

> The **SIGILL** and **SIGTERM** signals are not generated under Windows.

src/flag.rs
qnighy: I actually didn't know the difference between `raise` and `kill(getpid)`. I'll unify them as you said.

src/lib.rs
qnighy: In Windows SDK 10, they're defined in `C:\Program Files (x86)\Windows Kits\10\Include\%VERSION%\ucrt\signal.h` without `#ifdef` conditionals. So I guess they're uniform between architectures. I'm in favor of defining `SIGABRT_COMPAT`/`SIGBREAK` here and backport to `libc` if we can.

tests/iterator.rs
qnighy: This isn't `src/iterator.rs`, but `tests/iterator.rs`.

signal-hook-registry/src/lib.rs
qnighy: `signal()` comes before RCU store -- so I guess nothing happens (neither the default handler nor the registered handler is called) instead.

vorner: I'm not sure I understand. Or, are we both thinking about the same race condition?

This is what I fear might happen:

* SIGTERM (or something else) arrives. This resets the signal to default, which is terminate.
* It calls the handler.
* Before the handler can re-establish itself back from the default, another SIGTERM arrives. This one terminates the program, because it didn't get re-established.

It is probably quite rare for the user to press CTRL+C *that* fast, so it most probably won't happen, but in theory something could send them programmatically?

But I'm not sure about the theory.

signal-hook-registry/src/lib.rs
vorner: Ah, I see. That makes sense.

tests/iterator.rs
vorner: I see, oversight on my side :-). It makes sense.

signal-hook-registry/src/lib.rs
vorner: No. I'm just worrying that with SIGUSR1 and SIGUSR2 being the same signal in windows tests (as you define them), one test might somehow (not sure how) could make the other test fail by accident.

I'm proposing to keep the code as it is, just add a note with this worry, so *if* the problem ever happens, people will just try re-running it and not invest hours trying to figure out what will have happened.

vorner:
Getting there :-). There are these little details I'm not sure about I'd like you to have a look at, but if you think I'm wrong, it'd be fine to merge it as it is.

I'm thinking about one other thing. This sub-crate is already 1.0. Adding new functions is fine, but that means promising not to ever remove or change their signatures. I'll probably ask around a bit, if there's a convention how to get an unstable part of API in otherwise stable crate ‒ so maybe I'll hold releasing it a bit after it gets merged, to be on the safe side. I hope it's not a problem. If you have a suggestion how to solve this, I'd be interested in hearing it.

signal-hook-registry/src/lib.rs
qnighy: Yeah I imagined a different one than yours, so I added another bullet here to describe yours. The one in my mind was:

- The program tries to register a handler for SIGINT. It acquires the RCU lock. We don't block signals here because there's no such machinery in Windows.
- `signal` is called. Our global `handler` function is registered.
- SIGINT arrives. As `handler` is already registered, the default handler isn't invoked. However, we haven't stored the RCU result yet so there's no shared handler registered.
- RCU store happens.

In this case we miss SIGINT where we expect the default handler ("before handler registration" case) or the registered handler ("after handler registration" case) to be invoked.

vorner:
Thank you. I'll mark the new functions/the ones that have different signature in some way to clearly state they may still change a bit and will release the registry crate over the weekend.

vorner:
Looking through the newly introduced API, there seems to be nearly nothing (only the `register_signal_unchecked` function), so I think it's fine to promise API stability on this.

Do you want to add yourself to the crate authors, before I release?

qnighy:
> Do you want to add yourself to the crate authors, before I release?

Yes, I'll prepare a PR in minutes...

PR #20, opened 2019-07-27
Add myself (@qnighy) to the list of authors.

qnighy:
I'd like to add myself to the list of authors to reflect #19.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "signal-hook"
 version = "0.1.10"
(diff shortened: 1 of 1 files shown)


vorner:
This is the wrong `Cargo.toml` (I'm releasing signal-hook-registry, as that one has the changes, so the release would not list you on crates.io). This is fine, though, I've copied the name and email from here to the correct file :-).

I'll of course include you in this Cargo.toml too once the rest of the changes are merged and they get released.

PR #22, opened 2019-09-17
Document what the pipe module writes to the file

datto-aparrill:
Mention that the pipe module writes one byte of garbage, so that the reader knows what to expect.

### DIFF
--- a/src/pipe.rs
+++ b/src/pipe.rs
@@ -1,8 +1,8 @@
 //! Module with the self-pipe pattern.
 //!
 //! One of the common patterns around signals is to have a pipe with both ends in the same program.
-//! Whenever there's a signal, the signal handler writes to the write end. The application then can
-//! handle the read end.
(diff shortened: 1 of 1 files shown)


src/pipe.rs
```diff
@@ -1,8 +1,8 @@
 //! Module with the self-pipe pattern.
 //!
 //! One of the common patterns around signals is to have a pipe with both ends in the same program.
-//! Whenever there's a signal, the signal handler writes to the write end. The application then can
-//! handle the read end.
+//! Whenever there's a signal, the signal handler writes one byte of garbage data to the write end.
```
vorner: Suggestion: add a note like `(unless the pipe is already full)`

vorner:
Rebasing onto master should fix CI in your branch now.

vorner:
@datto-aparrill Are you looking into that ↑, or should I just take it over?

Issue #24, opened 2019-10-13
Tokio 0.2 support


vorner:
Tokio is probably close to releasing its version 0.2 and this crate should be ready for it. Maybe it is even possible to do it in a way that one feature turns on support for tokio 0.1, while another support for tokio 0.2.

The implementation needs to be against the alpha releases for now, so it probably should stay in a branch for now, to try it out, but this can be ready for just setting the right dependency and releasing.

Note that supporting async-std is currently impossible, as there's no way to plug arbitrary file descriptors into it right now.

jjakobh:
I am currently trying to do this and it compiles but my tests fail with
`called Result::unwrap() on an Err value: Custom { kind: Other, error: "no current reactor" }`
for 
```rust
let mut runtime = Runtime::new().unwrap();
...
runtime.block_on(...)
```
and ` 'default Tokio reactor already set for execution context`
for
```rust
let reactor = tokio_net::driver::Reactor::new().unwrap();
let handle = reactor.handle();
let _guard = tokio_net::driver::set_default(&handle);
let mut runtime = Runtime::new().unwrap();

```

Any idea why this is happening?

vorner:
I think the first should work, if you're using runtime from the same version as the network primitives (AFAIK currently tokio doesn't yet have a migration path ready, so spawning 0.2 TcpStream or anything else onto tokio 0.1 reactor doesn't work).

The second one can't work, because Runtime also sets a reactor and these two fight each other.

Anyway, it's hard to tell from this little snippet. If you could maybe share the whole code you have, I might be able to help more.

jjakobh:
[This](https://github.com/jjakobh/signal-hook/tree/tokio-0.2) is my branch, the tests that fail are [tokio::repeated](https://github.com/jjakobh/signal-hook/blob/tokio-0.2/tests/tokio.rs#L29) and [tokio::delayed](https://github.com/jjakobh/signal-hook/blob/tokio-0.2/tests/tokio.rs#L51), and the relevant Tokio stuff is [here](https://github.com/jjakobh/signal-hook/blob/tokio-0.2/src/iterator.rs#L541).

`rg tokio Cargo.lock` shows that there is only `tokio 0.2-alpha.6` (and `tokio-timer 0.3.0-alpha.6`) in the dependency tree, so AFAICT there shouldn't be any conflicts.

vorner:
[...]

PR #25, opened 2019-10-14
update to rust 2018

jjakobh:


vorner:
Hello

Can I ask for the motivation behind this?

The crate currently tries to support Rust 1.26 (which is the one in Debian). I know that one is old, but it doesn't really *need* any of the newer features. So if updating to Rust 2018 is just because of aesthetics, I don't think dropping the old support is worth it.

jjakobh:
I wanted to give updating `signal-hook` to Tokio 0.2 a try and do this before because I assumed Tokios MSRV would be higher, but now I cant find tokios policy on rust versions so I'm not sure where I got that idea from.

vorner:
That makes some sense. Though, I'd like to point out that tokio is only optional dependency. So even if support for tokio 0.2 is added, it is possible to support 1.26 without it enabled.

But I'm probably willing to talk about bumping the minimal version for signal-hook itself, especially if updating to tokio 0.2 requires a major version bump/API breackage. On the other hand, I'd still like to keep the signal-hook-registry compatible.

jjakobh:
sure, makes sense

vorner:
Oh, by the way, if you want to work on the tokio 0.2 support, I'd suggest you write a comment to #24 and claim it (or, I can just assign you there). Just to prevent duplicate effort. And any such effort in that direction will be appreciated :-).

PR #26, opened 2019-10-16
Tokio 0.2 support

jakobhellermann:
`cargo +1.26.2 check` compiles successfully but `cargo +1.26.2 test` doesn't, because apparently cargo doesn't support optional dev-dependencies.

`cargo +{beta,nightly} test --all-features` passes all the tests.

Should I adjust the readme to say that the standard MSRV 1.26 but for `tokio-support` it is https://github.com/tokio-rs/tokio#supported-rust-versions?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,17 @@ script:
       export RUST_BACKTRACE=1 &&
       export CARGO_INCREMENTAL=1 &&
       (test "$TRAVIS_RUST_VERSION" == 1.26.0 || rm -f Cargo.lock) &&
+      # for all targets on all platforms
       cargo build --package signal-hook --package signal-hook-registry &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo build --all --all-features) &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo test --package signal-hook --package signal-hook-registry) &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo test --all --all-features) &&
       cargo doc --no-deps &&
+      # for stable/beta/nightly on all platforms
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo test   --package signal-hook --package signal-hook-registry) &&
       (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo clippy --package signal-hook --package signal-hook-registry --tests -- --deny clippy::all) &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo clippy --package tokio-tests --tests -- --deny clippy::all) &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo fmt)
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo fmt) &&
+      # for beta/nightly on macOS/linux
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || test "$TRAVIS_OS_NAME" == "windows" || cargo build  --all --all-features) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || test "$TRAVIS_OS_NAME" == "windows" || cargo test   --all --all-features) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || test "$TRAVIS_OS_NAME" == "windows" || cargo clippy --package tokio-tests --tests -- --deny clippy::all)
 
 matrix:
     allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,13 @@ script:
       export RUST_BACKTRACE=1 &&
       export CARGO_INCREMENTAL=1 &&
       (test "$TRAVIS_RUST_VERSION" == 1.26.0 || rm -f Cargo.lock) &&
-      cargo build --all &&
+      cargo build --package signal-hook --package signal-hook-registry &&
       (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo build --all --all-features) &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo test --all) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo test --package signal-hook --package signal-hook-registry) &&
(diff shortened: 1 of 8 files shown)


vorner:
I'll have a look during the weekend at the latest.

I see the problem (though the CI fails even on beta & nightly for some reason, can you have a look?). I'll think if I can come up with some kind of workaround for this, it is unfortunate :-|.

.travis.yml
```diff
@@ -22,11 +22,11 @@ script:
...
-      cargo build --all --all-features &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo test --all) &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo test --all --all-features) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo build --all --all-features) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo test --all) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo test --all --all-features) &&
       cargo doc --no-deps &&
-      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo clippy --all --tests -- --deny clippy::all) &&
+      (test "$TRAVIS_RUST_VERSION" == 1.26.0 || test "$TRAVIS_RUST_VERSION" == "stable" || cargo clippy --all --tests -- --deny clippy::all) &&
       (test "$TRAVIS_RUST_VERSION" == 1.26.0 || cargo fmt)
```
vorner: I guess it makes sense to disable the tests on stable right now. But I'd like to wait for the 1.39 / tokio 0.2 release before actually putting this into master and by then we need to not forget re-enabling them.

Cargo.toml
```diff
@@ -24,15 +24,15 @@ members = [
...
+mio = { version = "0.6.19", optional = true }
 signal-hook-registry = { version = "~1", path = "signal-hook-registry" }
-tokio-reactor = { version = "~0.1", optional = true }
+tokio-net = { version = "0.2.0-alpha.6", optional = true }
 
 [dev-dependencies]
-version-sync = "~0.8"
-tokio = "~0.1"
+version-sync = "0.8.1"
+tokio = "0.2.0-alpha.6"
```
vorner: Why do you change the versions to not include the `~`? I like to include it because it is more readable. Without the sign, it is the same (requiring semver-compatible version), but this way it's clear it is not the *exact* version.

Furthermore, why do you increase the minimal required patch version? Do you really need at least version 0.8.1 of version-sync, etc?

signal-hook-registry/Cargo.toml
```diff
@@ -18,9 +18,9 @@ travis-ci = { repository = "vorner/signal-hook" }
...
 
 [dependencies]
-arc-swap = "~0.4"
-libc = "~0.2"
+arc-swap = "0.4.3"
+libc = "0.2.62"
 
 [dev-dependencies]
-version-sync = "~0.8"
+version-sync = "0.8.1"
```
vorner: (again, why the dependencies)

src/iterator.rs
```diff
@@ -593,25 +597,31 @@ mod tokio_support {
...
-        ///     let wait_signal = Signals::new(&[signal_hook::SIGUSR1])?
-        ///         .into_async()?
-        ///         .into_future()
-        ///         .map(|sig| assert_eq!(sig.0.unwrap(), signal_hook::SIGUSR1))
-        ///         .map_err(|e| panic!("{}", e.0));
+        ///     let mut runtime = Runtime::new().unwrap();
+        ///
+        ///     let mut signals = Signals::new(&[signal_hook::SIGUSR1])?
+        ///         .into_async()?;
+        ///     let wait_signal = signals
```
vorner: Did you split the method chain on purpose?

vorner:
Thinking about it more, I don't know, maybe having another copy and duplicating code is not that great an idea…

Cargo.toml
jjakobh: I think I ran `cargo upgrade` which seems to prefer this version format

src/iterator.rs
jjakobh: I did it because there was the error message about a temporary freed while in use, but I just tried unsplitting the chain and it worked

jjakobh:
alright, now the following is tested:
 - 1.26.2:
   - `cargo build` for `signal-hook` and `signal-hook-registry`
- stable:
   - `cargo {build,test,clippy}` for `signal-hook` and `signal-hook-registry`
- nightly:
   - `cargo {build,test,clippy} --all --all-features`

I'm sure travis will still fail for some reason.

jjakobh:
So the subcrate for tokio works, but what do you want to do about the doctest?
Make it `rust,ignore` and put a `//TODO make sure it compiles` next to it?

src/iterator.rs
```diff
@@ -585,33 +589,38 @@ mod tokio_support {
         ///
         /// # Examples
         ///
-        /// ```rust
+        /// ```rust,norun
```
vorner: Why is this norun? Won't it just not compile at all if we don't have the dependency anyway?

tokio-tests/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #![cfg(not(windows))]
-
-#[cfg(feature = "tokio-support")]
+#[cfg(test)]
 mod tests {
```
vorner: I think that now that the test is in a separate crate and run selectively, the whole `cfg` and module thing is no longer needed.

src/iterator.rs
jjakobh: I changed this to `rust,ignore` locally but forgot to push it.

tokio-tests/src/lib.rs
jjakobh: I'm not sure how this is usually done but I think the `mod tests` is still useful because else all the imports will emit an unused import warning.

vorner: You're right there. I was thinking in the line of files living in `tests`, which don't need this as they are always being tests.

vorner:
It seems there's still one more CI failure. This one looks related to Windows not having much support yet, so large parts are simply completely missing ‒ including the iterator.

vorner:
I've made a copy of the branch in this repository. I want to wait with merging until tokio-0.2 is actually out. As it is, with removal of tokio-0.1 support, the version will have to be incremented. It would have to be incremented again once tokio goes non-alpha. And I want to be able to release fixes from master if there's a reason in the meantime.

jjakobh:
Ok, sure

vorner:
I've just noticed tokio 0.2 is out. I'll be merging and releasing this soon (probably over the weekend, I don't know if I'll get to it before that).

jakobhellermann:
well, it seems like first the tokio support has to be rewritten using the new [`Registration`](https://docs.rs/tokio/0.2.1/tokio/io/struct.Registration.html) which I don't think I'm able to do myself

vorner:
Ah, that's a bit sad. Anyway, I'll have a look and see over the weekend if I can make it work without too many changes.

Alternatively, tokio has signal support itself (which sits on top of signal-hook-registry), so in theory it would be fine to just drop the support and lead the people to their implementation.

vorner:
I've played with it a bit and it seems to work. I still need to update documentation and examples and stuff and I've run out of energy for today, but if you're interested, you can have a look at https://github.com/vorner/signal-hook/commit/16402edb4c14d4c01172d8cb86d6e1f1d879844b. The addition of dev-dependencies is just temporary measure, until I verify it works, then I'll make it ignored again.

PR #31, opened 2020-01-20
lib: add SIGINFO

grimreaper:
modern operating systems have an additional non-POSIX signal SIGINFO.
This is typically sent to an application by pressing ctrl-t.

### DIFF
--- a/tests/iterator.rs
+++ b/tests/iterator.rs
@@ -67,3 +67,10 @@ fn signals_block_wait() {
         .expect_err("Wait didn't wait properly");
     assert_eq!(err, RecvTimeoutError::Timeout);
 }
+
+#[cfg(not(any(target_os = "linux", windows)))]
+use signal_hook::SIGINFO;
+#[test]
+fn siginfo_is_valid_signal() {
+    let _signal: Signals = Signals::new(&[SIGINFO]).unwrap();
+}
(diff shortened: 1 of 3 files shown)


grimreaper:
FWIW I am fairly new to (modern) rust. Please let me know if there a way to improve this.

tests/iterator.rs
```diff
@@ -67,3 +67,10 @@ fn signals_block_wait() {
         .expect_err("Wait didn't wait properly");
     assert_eq!(err, RecvTimeoutError::Timeout);
 }
+
+#[cfg(not(any(target_os = "linux", windows)))]
+use signal_hook::SIGINFO;
+#[test]
```
vorner: Is this test iterator related? I'd say this belongs somewhere slightly differently than tests/iterator.

Style:
* I prefer to have all uses together at the top.
* There should be a space above `#[test]` to visually separate the two items.

Also note that the CI fails on linux because of this, you omit importing the name, but you don't omit the test using it itself.

grimreaper:
> Would it alternatively make sense to simply document the „Works with other signals too, use libc::SIG_WHATEVER“ directly?

Yep. I'll close this PR and re-open one that says exactly this. I did not realize that this would be the case.

PR #32, opened 2020-01-20
lib: document that any signal value can be used

grimreaper:
When initially reading the documentation for this library I was unaware
that this was legal. Document this to reduce confusion.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,6 @@
 #![deny(missing_docs, warnings)]
 // Don't fail on links to things not enabled in features
 #![allow(unknown_lints, intra_doc_link_resolution_failure)]
-
 //! Library for easier and safe Unix signal handling
 //!
 //! Unix signals are inherently hard to handle correctly, for several reasons:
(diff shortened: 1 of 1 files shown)


grimreaper:
that makes sense! I'll go ahead and try to do so.

grimreaper:
What do you think of this example?

src/lib.rs
```diff
@@ -85,6 +85,15 @@
...
 //! exception of Windows.
 //!
+//! Non-standard signals are also supported. Pass the signal value directly from `libc` or use
+//! the numeric value directly.
+//! ```rust
+//!    use std::sync::Arc;
+//!    use std::sync::atomic::{AtomicBool};
+//!    let term = Arc::new(AtomicBool::new(false));
+//!    let _ = signal_hook::flag::register(libc::SIGINT, Arc::clone(&term));
+//! ```
```
vorner: Just a small style nitpick:

* Do not indent inside the example. The rendered docs then contain the indentation too and it looks odd. There's no outer block to indent against.
* Please leave a newline between the paragraph and the example at the top.

grimreaper: Excellent. I also figured out how to generate the documentation so now it looks like this:
image

vorner:
Thanks. I don't have any more changes to do right now, so I'm going to release this once the tests and everything pass on master.

PR #34, opened 2020-04-22
Bump to mio 0.7.0

chrisduerr:


vorner:
Hello

Thank you for taking the interest and sending the pull request.

Nevertheless, I'm afraid this is an API breaking change, so this would need bumping the version. But that would make it a semver-incompatible update to even people not using the mio feature.

Considering I'm also thinking about going to 1.0 on signal-hook and mio is not yet 1.0, I'm wondering if the right way forward isn't extracting this into a separate crate (`signal-hook-mio`) so it can be versioned separately. I think it could still live in this repository as a sub-crate (it already contains `signal-hook-registry`).

Would you like to give the extraction a try, or should I take it over from here? (Even then, figuring how to rename stuff around the mio update is valuable and won't get lost, so thanks for it).

vorner:
Alternatively, one could also add a `mio-0_7` feature flag, duplicate the support and support both in parallel. That doesn't sound very clean solution, though.

vorner:
Also, the tests fail currently: https://travis-ci.org/github/vorner/signal-hook/builds/678396874

chrisduerr:
> I'm wondering if the right way forward isn't extracting this into a separate crate (signal-hook-mio) so it can be versioned separately.

How would you approach that? Just creating a wrapper around the `Signal` type and implementing everything on top of that? Sounds like it could be pretty annoying.

vorner:
Oh, it's not a separate type like with the tokio async thing. That makes it harder :-|. A wrapper would be annoying indeed.

Let me think about it for a while if there's a good solution.

chrisduerr:
> Let me think about it for a while if there's a good solution.

Sounds good to me, there's no rush with this PR since basically nobody has updated to mio yet because of the massive breaking API changes.

If there's another change like this coming I could definitely see why people wouldn't want to expose that as public API in a 1.0+ crate, though the integration in signal-hook is fairly straight-forward since nothing really is exposed.

vorner:
I haven't thought of anything great. So I'd say the best solution for now is to add another feature (eg. `mio-0_7-support`) and duplicate the code.

Maybe some kind of macro could be used to generate both without that much duplication, as it seems the changes are mostly just renaming things and these could be passed to it. Will you update the PR?

I hope to find a better solution before declaring it 1.0 (which won't happen right now).

chrisduerr:
Shipping both mio 0.7 and 0.6 seems like a strange choice. I'm not sure how many people that would benefit and I'd assume that finding that feature alone would be tough.

vorner:
I wouldn't expect a single person to activate both features. But I suspect some people might want to keep using the 0.6 for a while and some would like to migrate to 0.7, so they would activate just that one relevant feature. I don't want to make a 0.2 version, add some nice features and leave the mio-0.6 people stranded.

I don't think finding that feature would be more difficult than it is finding the `mio-support` feature now. It's in documentation and in `Cargo.toml` and one needs to find it either way.

chrisduerr:
Should now both work at the same time.

Cargo.toml
```diff
@@ -15,7 +15,8 @@ maintenance = { status = "actively-developed" }
 
 [features]
 mio-support = ["mio"]
-tokio-support = ["futures", "mio-support", "tokio-reactor"]
+mio-0_7-support = ["mio_0_7"]
+tokio-support = ["futures", "mio-support", "mio-0_7-support", "tokio-reactor"]
```
vorner, line 19: I don't believe tokio needs both.

src/iterator.rs
```diff
@@ -138,9 +138,10 @@ impl Drop for RegisteredSignals {
 ///
 /// # `mio` support
 ///
-/// If the crate is compiled with the `mio-support` flag, the `Signals` becomes pluggable into
-/// `mio` (it implements the `Source` trait). If it becomes readable, there may be new signals to
-/// pick up. The structure is expected to be registered with level triggered mode.
+/// If the crate is compiled with the `mio-support` or `mio-0_7-support` flags, the `Signals`
+/// becomes pluggable into `mio` (it implements the `Source` trait). If it becomes readable, there
```
vorner, line 142: Maybe extend it to something like „pluggable into `mio` version `0.6` or `0.7` respectively“

chrisduerr:
Should be all good now.

vorner:
Thank you. I've made a release with it.

PR #37, opened 2020-06-13
Smol support (and indirect async_std 1.6+ support)

tfkhim:
Hi,

I'd like to submit some code for supporting the [smol runtime](https://docs.rs/crate/smol/). I tried to be a bit more generic than necessary by separating the implementation into a smol specific part and a futures 0.3 part. The later could be reused for integrating with other runtimes that allow getting an AsyncRead instance through registering the read fd with it.

# Possible improvement

The implementation of the poll_next method is mostly a copy and paste from the poll method for tokio-support. I believe the introducing a helper struct containing the signals and position members offering methods to advance and reset the iteration may be a possibility to reduce duplication.

But I'm not sure if it is worth doing. Saved lines of code would be minimal and the complexity of the code isn't very high anyway.

# async_std support through the backdoor

Since version 1.6.0 async_std also uses smol in the background. So this change allows the Signals iterator to be used in async_std code, too. But this support could break at any time because using smol should be considered as an implementation detail of async_std and there is (as far as I can see) no official API to register a file descriptor with its runtime.

Therefore I didn't add any documentation for this yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-task"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "blocking"
+version = "0.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "futures-channel 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "futures-util 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "once_cell 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "parking 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "waker-fn 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "byteorder"
@@ -41,6 +58,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cache-padded"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "cc"
+version = "1.0.55"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "cfg-if"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -51,6 +78,14 @@ version = "0.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "concurrent-queue"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cache-padded 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


vorner:
Hello

Thank you for the submission. I'll try to find the time to look at it soon. Optimistic estimate is during next 24 hours, pessimistic is something like a week.

vorner:
OK, I'm reading through the code. The code itself looks high-quality and nice.

However, I have few higher level worries here.

One of them is the duplication you mention above. I don't really care much about the saved lines, and agree that the actual *code* is simple enough. However, I feel like the *logic* and *correctness proof* behind the code is far from simple. Therefore, I don't really want to have multiple copies of it.

The second worry is about some kind of async runtime adapter proliferation through the code. When I started, there was just tokio 0.1 + mio 0.6. Now, while tokio 0.2 has its own signal support (based on signal-hook-registry), the code already got mio 0.7 support and now smol 0.1. I *guess* there'll eventually be smol 0.2, smol 0.3 and eventually something like smol 1.0. Each of them would have its own adapter and I don't really like removing APIs once they appear. This leads to API pollution, code bloat and something like good name exhaustion (I really don't like the addition of `into_stream` sitting next to `into_async` which kind of do the same but with different runtimes and it's absolutely not clear which runtime which one belongs to).

Furthermore, I'd really like to go 1.0 eventually and probably before all the async runtimes do, so optimally the crate would not expose any of *their* APIs by then.

So I'd like to explore a different approach. The idea is, this crate would provide some kind of adapter that would have the equivalent of `poll/poll_next`, ability to pull the file descriptor out, etc, but no direct support for any specific runtime. Then there would be other crates like `signal-hook-smol` that would use this exposed API of the adapter to wrap it into the smol-specific thing. These crates could happily live here in this repository, but they could have their own versioning schemes and allow for having a version for smol 0.1, smol 0.2 living in parallel, etc.

I don't know if you're in hurry for this feature. I see two possible ways forward. One of them is that this gets merged as is (well, with the de-duplication of code in some nice way, if possible) with the expectation that in maybe 0.2 the API will disappear (move to its own crate). Or that the exploration for the split happens now and this gets merged into the „new“ way. What would you prefer? Do you want to give a shot at the splitting/abstracting the adapter away?

stjepang:
As author of smol, I don't think this is necessary because you can already use signal-hook with it: https://github.com/stjepang/smol/blob/d43f850050d65de7536d9ccb19a0e3b4a3b8a679/examples/unix-signal.rs

I'm also holding a bit stronger opinion: libraries should not depend on runtimes nor should they inflict the choice of a runtime onto the user, period. In fact, if a library depends on a runtime, I simply won't use it.

vorner:
Looking at the example, I'd say the `Signals` abstraction is a bit higher-level and potentially more convenient when receiving multiple different signals. So while not strictly *necessary*, I'd say being able to use `Signals` as a stream instead of iterator is a nice to have.

> I'm also holding a bit stronger opinion: libraries should not depend on runtimes nor should they inflict the choice of a runtime onto the user, period.

Certainly a hard dependency on runtime would be a no-go for signal-hook. Partly because it's not even primarily async library to start with. Nevertheless, I do find it useful if the user can enable a feature or bring in another library that provides a more convenient integration into specific runtime. What I tried to sketch above is, the base signal-hook doesn't know anything about runtimes or async. You could use it manually as you show in the example, or you could bring in `signal-hook-smol` to get rid of the need to know about single bytes meaning something, etc and get something that „just works“. Or you could bring in `signal-hook-tokio` or `signal-hook-whatever-else` to bring in support for whatever else. While I don't want to inflict the choice onto the user, I think if the sole purpose of a library is to make something work well together with specific runtime, then it's fine to depend on that runtime.

tfkhim:
I agree with your worries about API pollution and a growing number of xyz-support feature flags. I assumed based on the existing flags that this is the solution you settled with. But focusing on the blocking style in the core crate and moving everything async related to small integration crates is more reasonable.

I'm not really in a hurry. I was just toying around with async-std and async-tungstenite and thought others may benefit from some of the code, too. 

I would like to make an attempt to split the async features apart. I won't have much time this week. But I'll try to wrap my head around it. When I have a more specific idea I'll open an issue to discuss it.

vorner:
> I'd like to make an attempt to split the async features apart. But I won't have much time this week. I'll try to wrap my head around it. If I have a more specific idea I'll open an issue to discuss it.

No hurry. I guess removal of these async things into its own crate will be a breaking change and polishing it out will take some time anyway.

Feel free to ask or discuss whatever :-).

I'll probably leave this PR open for now, as a reminder that something else will supersede it eventually.

vorner:
I believe this one got outdated by the other pull requests, so I'm closing it as a cleanup.

PR #44, opened 2020-07-31
Fix a bug in the signals_close_forever test

tfkhim:
Hi,

I found a small bug in the signals_close_forever test. The threads spawn logic wasn't executed until the final
for loop and one thread was spawned after the other because of the join in the for loop body. So there was never any parallelism in this test case.

### DIFF
--- a/tests/iterator.rs
+++ b/tests/iterator.rs
@@ -30,6 +30,10 @@ fn signals_close_forever() {
         })
     });
 
+    // The map method is lazy and will execute the function only during iteration. So we
+    // collect the JoinHandles to ensure the threads are actually started.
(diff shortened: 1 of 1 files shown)


vorner:
I think I don't understand what you mean. How does this change in any way change the semantics of the code? In both versions, the join handles just get dropped, nobody joins them and:

> A JoinHandle detaches the associated thread when it is dropped, which means that there is no longer any handle to thread and no way to join on it.

https://doc.rust-lang.org/std/thread/struct.JoinHandle.html

Or am I not seeing something important?

vorner:
Ah, sorry, scratch the previous, didn't read the comment in the code 😇 (I'm doing multiple things at once). Yes, you're right.

PR #45, opened 2020-08-04
pipe::register_raw: Take ownership, don't dup

### LINKED ISSUE #41: `pipe::register_raw` duplicate fd code doesn't behave as intended
`pipe::register_raw` duplicates the given file descriptor with `dup(2)`, and has a comment saying:

```
    // We want to set the FD non-blocking. But it belongs to the caller. Therefore, we make our own
    // copy with `dup` to play on instead.
```

I think there are two problems here:

1. The O_NONBLOCK flag is a *file status* flag, not a *file descriptor* flag. This means the flag is associated with the underlying object not with the fd that you used. So the reason given in the comment doesn't actually work; subsequent accesses via the original fd will also be non-blocking. I'm afraid I don't have a truly canonical reference (it's probably somewhere in some POSIX specification that I don't have time to find) but the GNU libc documentation about [duplicating descriptors](https://www.gnu.org/software/libc/manual/html_node/Duplicating-Descriptors.html#Duplicating-Descriptors) is pretty clear:
   > You can duplicate a file descriptor, or allocate another file descriptor that refers to the same open file as the original. Duplicate descriptors share one file position and one set of file status flags (see [File Status Flags](https://www.gnu.org/software/libc/manual/html_node/File-Status-Flags.html#File-Status-Flags)), but each has its own set of file descriptor flags (see [Descriptor Flags](https://www.gnu.org/software/libc/manual/html_node/Descriptor-Flags.html#Descriptor-Flags)).

   If you follow the [File Status Flags](https://www.gnu.org/software/libc/manual/html_node/File-Status-Flags.html#File-Status-Flags) link and from there the page about [Operating Modes](https://www.gnu.org/software/libc/manual/html_node/Operating-Modes.html#Operating-Modes) you'll see O_NONBLOCK is a file status flag, and the Descriptor Flags page says `O_CLOEXEC` is currently the *only* file descriptor flag.

2. File descriptors created with `dup(2)` start off with O_CLOEXEC *cleared* (regardless of whether the original fd had O_CLOEXEC set). I'm not an authority but I think best practice is to create all descriptors with O_CLOEXEC (if on a system that provides means to do that...) unless they are being deliberately set up to be passed to a child process. Otherwise child processes will end up with extra fds that give them access to things they shouldn't have access to. This is probably harmless in the common case that children are not malicious or extremely buggy, but still it's nicer to prevent it. `fcntl` `F_DUPFD_CLOEXEC` command can be used to duplicate a file descriptor with O_CLOEXEC set (atomically; ie, there is no point at which the new descriptor is missing the flag), or on systems that don't support `F_DUPFD_CLOEXEC` you can separately set O_CLOEXEC after `dup` (though technically there's a hole where if you fork from another thread the child will inherit the descriptor). Here is how fds are duplicated in the standard library: https://github.com/rust-lang/rust/blob/47ea6d90b073ab977cf072e2f5f46d63de532cc6/src/libstd/sys/unix/fd.rs#L223

I don't think these problems are hugely important but it would be nice to fix the O_CLOEXEC thing at least, and that should be fixable without an API change. For the O_NONBLOCK thing perhaps the documentation could be updated to warn that the underlying object will be set to non-blocking mode (I don't really see a better way there). With an API change I think it might be cleaner if register_raw takes ownership of the file descriptor, and the caller can do their own fd duplication if they want to retain direct access.

vorner:
Take ownership of the pipe. If the caller wants to dup it, they can.
Don't dup internally.

Set the O_NONBLOCK only if it can't be used with `send`.

Closes #41.

@johnbartholomew Could I ask you for second pair of eyes here?

### DIFF
--- a/src/pipe.rs
+++ b/src/pipe.rs
@@ -74,16 +74,25 @@
 //!     Ok(())
 //! }
 
-use std::io::Error;
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::io::{Error, ErrorKind};
+use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
 
 use libc::{self, c_int};
 
 use crate::SigId;
 
-struct OwnedFd(RawFd);
+#[derive(Copy, Clone)]
+pub(crate) enum WakeMethod {
+    Send,
+    Write,
+}
 
-impl OwnedFd {
+struct WakeFd {
+    fd: RawFd,
+    method: WakeMethod,
+}
+
+impl WakeFd {
     /// Sets close on exec and nonblock on the inner file descriptor.
     fn set_flags(&self) -> Result<(), Error> {
         unsafe {
@@ -98,23 +107,26 @@ impl OwnedFd {
         }
         Ok(())
     }
-}
-
-impl AsRawFd for OwnedFd {
-    fn as_raw_fd(&self) -> RawFd {
-        self.0
+    fn wake(&self) {
+        wake(self.fd, self.method);
     }
 }
 
-impl Drop for OwnedFd {
+impl AsRawFd for WakeFd {
+    fn as_raw_fd(&self) -> RawFd {
+        self.fd
+    }
+}
+
+impl Drop for WakeFd {
     fn drop(&mut self) {
         unsafe {
-            libc::close(self.0);
+            libc::close(self.fd);
         }
     }
 }
 
-pub(crate) fn wake(pipe: RawFd) {
+pub(crate) fn wake(pipe: RawFd, method: WakeMethod) {
     unsafe {
         // This writes some data into the pipe.
         //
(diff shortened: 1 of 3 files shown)


johnbartholomew:
Sorry, just got back from a break. I will try to take a look this week.

vorner:
No hurry, thank you :-)

src/pipe.rs
```diff
@@ -126,28 +138,54 @@ pub(crate) fn wake(pipe: RawFd) {
...
+        };
     }
 }
 
 /// Registers a write to a self-pipe whenever there's the signal.
 ///
-/// In this case, the pipe is taken as the `RawFd`. It is still the caller's responsibility to
-/// close it.
+/// In this case, the pipe is taken as the `RawFd`. It'll be closed on deregistration. Effectively,
+/// the funcion takes ownership of the file descriptor. This includes feeling free to set arbitrary
```
johnbartholomew, line 152: Typo s/funcion/function/

johnbartholomew:
Seems fine - I didn't notice anything wrong except one tiny typo :+1:

vorner:
Thanks :-)

PR #48, opened 2020-10-10
Add some tests for the Signals struct and fix a bug that showed up

tfkhim:
Because I added some more test cases for the Signal struct I found a bug in my changes for async crate split. See the commit message for some more details on this.

I should have added those test cases right before I started with the recent refactorings. Sorry for that :-(

### DIFF
--- a/tests/iterator.rs
+++ b/tests/iterator.rs
@@ -13,6 +13,22 @@ use signal_hook::{SIGUSR1, SIGUSR2};
 
 use serial_test::serial;
 
+fn send_sigusr1() {
+    unsafe { libc::raise(SIGUSR1) };
+}
+
+fn send_sigusr2() {
+    unsafe { libc::raise(SIGUSR2) };
+}
+
+fn setup_without_any_signals() -> Signals {
+    Signals::new(&[]).unwrap()
+}
+
+fn setup_for_sigusr2() -> Signals {
+    Signals::new(&[SIGUSR2]).unwrap()
+}
+
 #[test]
 #[serial]
 fn signals_close_forever() {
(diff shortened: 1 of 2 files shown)


tests/iterator.rs
```diff
@@ -75,3 +91,122 @@ fn signals_block_wait() {
...
+    let signals = setup_for_sigusr2();
+    signals.add_signal(SIGUSR1).unwrap();
+
+    send_sigusr1();
+    let mut recieved_signals = signals.wait();
+    send_sigusr2();
+
+    assert_eq!(recieved_signals.next(), Some(SIGUSR1));
+    assert_eq!(recieved_signals.next(), Some(SIGUSR2));
+    assert_eq!(recieved_signals.next(), None);
```
vorner, line 188: A little warning: the actual returned order is internal implementation detail and is unspecified. Collect to `HashSet` to compare? Or collect to vec and sort?

tests/iterator.rs
```diff
@@ -75,3 +91,122 @@ fn signals_block_wait() {
         .expect_err("Wait didn't wait properly");
     assert_eq!(err, RecvTimeoutError::Timeout);
 }
+
+#[test]
+#[serial]
+fn pending_doesnt_block_and_is_empty_if_no_signals_recieved() {
```
vorner, line 97: nitpick: The name is quite a mouthful. I sometimes pick shorter names and put a doc comment onto the function (not that it would render in docs, but it has a feeling of being the documentation of the purpose of the thing as opposed to some random comment around it).

tests/iterator.rs
```diff
@@ -75,3 +91,122 @@ fn signals_block_wait() {
...
+#[test]
+#[serial]
+fn wait_returns_recieved_signals() {
+    let signals = setup_for_sigusr2();
+    send_sigusr2();
+
+    let mut recieved_signals = signals.wait();
+
+    assert_eq!(recieved_signals.next(), Some(SIGUSR2));
+    assert_eq!(recieved_signals.next(), None);
```
vorner, line 114: I think it would be more convenient/readable to use something like `assert_eq(vec![SIGUSR2], received_signals.collect())` here and at other places. It is a bit less performant (allocating to the vec), but that shouldn't be a problem for tests.

vorner (commented):
Thank you for the tests and bugfix. And don't worry, not having enough tests is probably the standard in the software world and one can always say there should have been more from the beginning. You could say these tests should have existed in the crate from long before you came around to it O:-).

There are some tiny things, if you'd have a look at them.

tfkhim:
I picked the first part of the long test name because the test is focused around the non-blocking claim. Actually the second part of the name was just one assert to check the claim holds true.

I also added two macro rules to make the assertions more readable and remove the dependency on the internal signal ordering.

Can you squash both commits during the merge or should I do this in advance?

vorner:
Thanks :-). Yes, I can squash.

PR #51, opened 2020-11-11
Tokio 0.3 and async-std 1.0 support

tfkhim:
Eventually I'm able to implement the async-std support and also Tokio 0.3 support.

Kind of funny that this was the goal which started my signal-hook journey back in june :)

I added some macros to make common code for Tokio and async-std reusable. At the moment those live in their own small utility crate. Maybe there is a better place for them?


vorner (commented):
:thinking: The macro crate is a bit strange, I wonder if it makes sense to be separate. How big a chance there is someone might want to actually use it directly? Or, if someone would have both these crates in the dependency graph, so the code would be shared?

If not, I was wondering if we could just have a file in the repository (maybe in a `/shared` folder) and symlink it to all the relevant crates.

Otherwise, it looks nice :+1: 

Sorry that I haven't released the previous changes yet. Another thing that might be a breaking change appeared, but it should be mostly small and I hope to implement it soon.

tfkhim:
Using a symbolic link is a nice idea. I'm also more comfortable if we can keep those macros private. You're right, we agreed to put support for major releases into the root of the crate. I changed those things.

I can squash the last two commits into the appropriate initial ones if you do not have any other objections.

shared/async_macros.rs
```diff
@@ -0,0 +1,68 @@
+macro_rules! wrap_signal_iterator {
```
vorner, line 1: Can I ask for some comment at the top of the file, explaining what it does, why it is in this strange place, the trick with symlinks, etc… once once someone finds it here, they could be a bit surprised.

vorner (commented):
I'd say I'm happy with this and you could squash, but the tests are failing. It seems windows doesn't like the trick with symlinks. I don't know if there's a reasonable trick for them. But do these crates work on windows at all?

tfkhim:
The adapter crates don't work on windows. But the failing step is `cargo fmt`. We could deactivate it in the CI script. But future contributors to the Windows parts would have the same problem. With this in mind I get the feeling that the symlink isn't a good solution.

I see the following options:

* Go with the initial macro helper crate. The create requires a lot of boilerplate for two rather small macros. But in the end using crates is the recommended way for sharing code.
* Moving the macros inside the main `signal-hook` crate. This shouldn't introduce any dependencies to the async runtimes or the `futures` crate. But the macros are strongly related to the async functionality. So it would feel odd to have them in this crate.
* Do not try to be as DRY as possible between crates but only within a single crate. So implement the `async_std` crate by hand and use some macro in the Tokio crate to reduce duplication for wrapping the `SignalIterator` type. This also allows to tailor the documentation to the specific runtime. At the moment I had to write very generic documentation.
* Move the adapter crates into their own Rust workspace and repository which then contains only Unix things. Using symbolic links there would be OK. Makes things more complicated w. r. t. CI, consistency and developer quality of life.

At the moment I would go with option 3. The introduced duplication can be easier removed than the public macros if someone has a better solution in the future.

vorner:
> At the moment I would go with option 3. The introduced duplication can be easier removed than the public macros if someone has a better solution in the future.

I agree the option 3, to duplicate a bit of code, is probably the least bad option, so lets go with it.

tfkhim:
I removed the symbolic link. Hopefully the tests are OK now.

tfkhim:
The remaining failure in the test also appears on the master branch. It looks like clippy got some additional tests recently.

I'm not sure if you already ran into this issue and fixed it in some other branch. If not I can add the suggested fix. Using `try_for_each` in this scenario looks correct to me.

If you do not have any further concerns I would squash the history into one commit for Tokio 0.3 and one for async-std and possibly add a third one for the clippy issue mentioned above.

vorner (approved):
Yes, I think you can squash / clean up the history, I'll be happy to merge it. I'm holding off some other branch I have ready so I don't make collisions for you, but I hope to do the release really soon after that 😇.

Ignore this particular failure for now, I'll take care of it. I don't know if I can use `try_for_each` or if it's newer than what rustc the crate currently tries to support, but I'll deal with it one way or another.

vorner:
Thank you :+1:

And I really promise to do some release soon.

I would like to accompany the release with a blog post and I would definitely like to thank you there and give you credit for all the work you've done. But I would like to check first, are you OK with that?

tfkhim:
> Thank you +1
> 
> And I really promise to do some release soon.
> 
> I would like to accompany the release with a blog post and I would definitely like to thank you there and give you credit for all the work you've done. But I would like to check first, are you OK with that?

It was my pleasure. I have to thank you for the good collaboration, too :+1: 

I will be honored if you mention me in your blog post :smiley:

PR #62, opened 2020-11-27
Add additional signal reexports

chrisduerr:
This adds the following signals to the `libc` reexports:

| Signal    | Description                            |
| --------- | -------------------------------------- |
| SIGTSTP   | Stop typed at terminal                 |
| SIGPWR    | Power failure (System V)               |
| SIGTTIN   | Terminal input for background process  |
| SIGTTOU   | Terminal output for background process |
| SIGURG    | Urgent condition on socket (4.2BSD)    |
| SIGVTALRM | Virtual alarm clock (4.2BSD)           |
| SIGXCPU   | CPU time limit exceeded (4.2BSD)       |
| SIGXFSZ   | File size limit exceeded (4.2BSD)      |

These are all the signals that are defined in `libc` which are not
aliases to other existing signals. The main useful one here is `SIGTSTP`
for handling ^Z, but adding all missing signals seems sensible.

The only other signal documented which I have not added is `SIGEMT`
(Emulator trap), since it is not exported by `libc`.


vorner:
Hello

Can you provide the motivation why these should be re-exported? I'm not really much of a fan of re-exporting in general and it kind of clutters the docs and API surface to have them all there. Note that as these are just constants, so you can always use them directly from `libc` and everything will just work.

Are you 100% sure these are available on all reasonable platforms? I'm afraid the CI does not (yet) cover everything.

chrisduerr:
> Can you provide the motivation why these should be re-exported? I'm not really much of a fan of re-exporting in general and it kind of clutters the docs and API surface to have them all there. Note that as these are just constants, so you can always use them directly from libc and everything will just work.

I use them from libc right now, but if the argument is that you don't need to reexport them, then `signal-hook` probably shouldn't reexport anything?

> Are you 100% sure these are available on all reasonable platforms? I'm afraid the CI does not (yet) cover everything.

I am not. But of the signals I have added only `SIGPWR` is not part of the POSIX standard. I've considered only adding the missing POSIX signals, but some of the existing signals already aren't part of that (see `SIGWINCH`/`SIGIO`). I suppose `SIGPWR` is the only non-POSIX signal that is from System V though.

vorner:
> I use them from libc right now, but if the argument is that you don't need to reexport them, then signal-hook probably shouldn't reexport anything?

That actually was my original stance. But I got swayed by users asking for them for comfort, so they don't have to put both signal-hook and libc into `Cargo.toml` and the current state is a bit of a compromise between having the common one there and cluttering the docs and API with every signal nobody ever used.

So mostly I'm asking if you think these would be used commonly enough to be included and the additional comfort for some people would outweight the discomfort of others that have to scroll through them in docs. That `SIGTSP` seems like a good candidate, but I do have some doubts about eg. `SIGXFSZ` ever being used in practice.

> I am not.

Give me a while, I have an idea how to add bunch of lightweight checks for a lot of platforms that I *think* are supported, now that I've migrated to github actions recently. Then we can rebase it on top of that and see.

As for them being POSIX… I'm specifically worried about android systems. That one doesn't seem to be fully posix compliant and there already are few hacks inside the code to deal with it.

chrisduerr:
> So mostly I'm asking if you think these would be used commonly enough to be included and the additional comfort for some people would outweight the discomfort of others that have to scroll through them in docs. That SIGTSP seems like a good candidate, but I do have some doubts about eg. SIGXFSZ ever being used in practice.

The only reason for this PR is `SIGTSTP`, I've not used any of the others and have no need to do so in the future. But effectively this PR will eliminate the need for any future PR doing the exact same thing again. If there are reexports, I'd say one might as well reexport everything? Leaving out just one or two seems somewhat strange?

> As for them being POSIX… I'm specifically worried about android systems. That one doesn't seem to be fully posix compliant and there already are few hacks inside the code to deal with it.

I have absolutely no clue at all about Android, so I can't help you there.

vorner:
> Leaving out just one or two seems somewhat strange?

That's true. But then, there are two questions:

* Should the crate be moving in the direction of covering them all, or removing the ones that already seem like not really that useful. I certainly missed the opportunity when releasing 0.2 :-|.
* What is, authoritatively, the complete list of signals on each platform? Not even the `libc` seems to have a good answer to that.

Anyway, do you think you could rebase this onto current master? I've added bunch of targets to be checked on the CI. If some of the signals are not supported on many of them, we somewhat easier life deciding about them.

chrisduerr:
> Should the crate be moving in the direction of covering them all, or removing the ones that already seem like not really that useful. I certainly missed the opportunity when releasing 0.2 :-|.

Given how few signals these are and how all of them are just reexports from `libc`, I think the only thing that makes sense is to support all of them or none of them. Supporting only the "important" ones is arbitrary (who decides what is important?) and not beneficial (reexporting 1 signal vs 10 doesn't really make things more difficult).

> What is, authoritatively, the complete list of signals on each platform? Not even the libc seems to have a good answer to that.

I'm not sure, all I've done is read `man 7 signal` and added every single signal that is in there which is exported by libc and not an alias for an existing signal. But this should definitely be a "complete" (except for `SIGEMT`) list of signals for Linux at least. If there are some BSD signals missing, someone can probably send the same PR for BSD? But I'd honestly assume that this PR covers most signals on Unix that are in use.

> Anyway, do you think you could rebase this onto current master? I've added bunch of targets to be checked on the CI. If some of the signals are not supported on many of them, we somewhat easier life deciding about them.

I mean I can't tell you what to do. Personally I'm fine with just adding `SIGTSTP`. But assuming this works on all platforms, I think it makes the most sense to add all of them. That way there are no arbitrary decisions, either use libc for everything or use signal_hook for everything, no need to mix and match (which is basically the worst solution possible).

vorner:
When I'm concidering #69 that might mandate some restructuring and therefore release of 0.3, eventually, I think I'm getting persuaded, in a way. I'd aim to eventually re-export all of them, but move them to a submodule, not to root.

I'll postpone merging this until I decide if I want to go about it by starting to throw breaking changes in, or if I want to start migrating through series of `#[doc(hidden)]` and `#[deprecated]` attributes. But I'll get back to it in few days.

PR #71, opened 2020-12-04
Do not consume the Signals instance in the forever method

### LINKED ISSUE #70: Signals::forever should not consume
There's no actual reason to consume, as repeatedly calling wait is possible on `&mut` reference. Also, consuming makes it hard to use after a `for` with `break` in it.

tfkhim:
This code should fix #70

This is a breaking change for the signal-hook crate but it isn't a breaking change for the adapter crates.

### DIFF
--- a/src/iterator/backend.rs
+++ b/src/iterator/backend.rs
@@ -17,7 +17,7 @@
 //! Use the [`Signals`][crate::iterator::Signals] struct or one of the types
 //! contained in the adapter libraries instead.
 
-use std::borrow::Borrow;
+use std::borrow::{Borrow, BorrowMut};
 use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::io::Error;
 use std::mem::MaybeUninit;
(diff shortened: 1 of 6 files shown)


src/iterator/backend.rs
```diff
@@ -414,19 +414,20 @@ pub enum PollResult<O> {
...
-{
+impl<SD, E: Exfiltrator> SignalIterator<SD, E> {
     /// Create a new infinite iterator for signals registered with the passed
     /// in [`SignalDelivery`] instance.
-    pub fn new(mut signals: SignalDelivery<R, E>) -> Self {
-        let iter = signals.pending();
+    pub fn new<R>(mut signals: SD) -> Self
+    where
+        SD: BorrowMut<SignalDelivery<R, E>>,
+        R: 'static + AsRawFd + Send + Sync,
```
vorner, line 428: I feel a bit uneasy about these trait bounds, for two reasons:

* I think `R` and the bounds should be on the impl block (if there are some methods that aren't influenced by it, then maybe having two impl blocks). Because `new` or `handle` or anything else is not influenced by `R`, really, so they should not have it in their signatures.
* Are we sure there's no situation at which the user might need to explicitly specify `R`?

vorner (commented):
Hello

Thank you. It mostly looks good (with that one nitpick).

> This is a breaking change for the signal-hook crate but it isn't a breaking change for the adapter crates.

As it is breaking for `signal-hook`, it'll have a version number bumped. And as the adaptor crates expose types from signal-hook in its public API (for example, Handle), it makes it breaking even for them :-(.

However, as I mentioned in #70, I still need to decide when and how I'm going to start putting the breaking changes in (I envision few others) and I'm not sure if I want them in master yet. I hope you won't mind if this sits here for a while (how long probably depends on the plan).

src/iterator/backend.rs
tfkhim: Moving the trait bounds to the impl isn't possible in the current state because this woudln't constrain the R parameter. I tried this first but found out it's not working. This issue is desribed [here](https://github.com/rust-lang/rust/issues/56209).

It would be possible to introducing a R parameter directly in the struct:

```
struct SignalIterator<R, SD, E>
where
    R: 'static + AsRawFd + Send + Sync,
    E: Exfiltrator,
    SD: BorrowMut<SignalDelivery<R, E>,
{
    std::marker::PhantomData<R>,
    ...
}

impl<R, SD, E> SignalIterator<R, SD, E>
where
    R: 'static + AsRawFd + Send + Sync,
    E: Exfiltrator,
    SD: BorrowMut<SignalDelivery<R, E>,
{
    ...
}
```

Then the R parameter is constraint in the impl and it is possible to add the other bounds there, too.

Can you elaborate your second point? I'm not sure if you are concerned about the missing possibility to explicitly specify it by the user or the user has to specify it in some rare cases.

vorner: I was worried that the user would have to specify it in some rare cases (or less rare ones).

But on the second look, this is the backend of the iterator and the only ones interacting with it are the authors of the async crates, so even if it was true, it would be fine. So let's leave it as it is.

PR #72, opened 2020-12-14
Remove futures dependency from async-std adapter

ooesili:
The `async-std` crate does not actually depend on the `futures` crate
for the types `Context`, `Poll`, `AsyncRead` (simply called `Read` in
`async-std`), `Stream`, and `StreamExt`. It instead relies on the
versions of `Context` and `Poll` from the Rust standard library, which
it re-exports, and provides its own definition of the `Read`, `Stream`,
and `StreamExt` traits. For this reason users of the
`signal-hook-async-std` crate are unlikely to depend on the `futures`
crate themselves, making this crate's dependency on `futures`
unnecessary.

This commit replaces these types with the versions from `async-std` and
removes the `futures` dependency from the `signal-hook-async-std` crate.

In another project I am working on that uses `signal-hook` and
`async-std`, this change removed the following 10 crates from my
dependency list:

- futures
- futures-executor
- futures-macro
- futures-sink
- futures-task
- futures-util
- pin-project
- pin-project-internal
- proc-macro-hack
- proc-macro-nested

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -915,7 +915,6 @@ name = "signal-hook-async-std"
 version = "0.1.0"
 dependencies = [
  "async-std",
- "futures 0.3.8",
  "libc",
  "serial_test",
  "signal-hook",
--- a/signal-hook-async-std/Cargo.toml
+++ b/signal-hook-async-std/Cargo.toml
@@ -22,7 +22,6 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 libc = "~0.2"
 async-std = "~1"
-futures = "~0.3"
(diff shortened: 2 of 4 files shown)


vorner:
Nice! That cutting down on dependencies is cool.

But I have a question. What is the relation between `futures::Stream` and `async-std::Stream`? I'm asking to know if this is a breaking change or not (so I know which version number to bump), because AFAIK it now implements a completely different trait, right?

vorner:
It doesn't matter. I'm doing some breaking changes anyway, so it'll have to be released with the bumped version.

PR #87, opened 2021-01-20
Update signal-hook-tokio for tokio v1.0

### LINKED ISSUE #83: Update signal-hook-tokio for tokio 1.0
At that point, we probably want to get rid of the support for the older `0.*` branches and remove the features of the crate.

bdash:
This drops support for pre-1.0 versions of tokio, and removes the per-version features and modules.

Fixes #83.


signal-hook-tokio/Cargo.toml
```diff
@@ -19,20 +19,14 @@ edition = "2018"
...
-tokio-0_1 = { package = "tokio", version = "~0.1", optional = true }
-futures-0_1 = { package = "futures", version = "~0.1", optional = true }
-tokio-0_3 = { package = "tokio", version = "~0.3", features = ["net"], optional = true }
-futures-0_3 = { package = "futures", version = "~0.3", optional = true }
+futures = "~0.3"
+tokio = { version = "~1.0", features = ["net"] }
 
 [dev-dependencies]
-tokio-0_3 = { package = "tokio", version = "~0.3", features = ["full"]}
+tokio = { package = "tokio", version = "~1.0", features = ["full"] }
```
vorner, line 29: I've recently discovered that this would disallow versions like `1.1.0`, which is probably not what we want. Can you update them to either `~1` or `^1.0`, which would allow them? (on both places here)

signal-hook-tokio/src/lib.rs
```diff
@@ -4,16 +4,57 @@
 
 //! A crate for integrating signal handling with the Tokio runtime.
 //!
-//! There are different sub modules for supporting different Tokio
-//! versions. The support for a version must be activated by a
-//! feature flag:
+//! This provides the [`Signals`][Signals] struct which acts as a
+//! [`Stream`][`futures::stream::Stream`] of signals.
```
vorner, line 8: :thinking: I'm bit unsure here. So we are going to „stabilize“ in the form we are going to support one true version of tokio & friends. But:

* There's tokio-stream at 0.1.
* We are using futures, but 0.3, not 1.0 (which also doesn't exist AFAIK).

I wonder which one of these we *should* be using or if we should support multiple by feature flags (by calling into some kind of `poll_next_inner`).

signal-hook-tokio/src/lib.rs
```diff
@@ -73,179 +114,37 @@ macro_rules! implement_signals_with_pipe {
...
-    use futures_0_1::stream::Stream;
-    use futures_0_1::{Async, Poll};
-    use signal_hook::iterator::backend::PollResult;
-    use tokio_0_1::io::AsyncRead;
-    use tokio_0_1::net::unix::UnixStream;
+use std::pin::Pin;
 
-    implement_signals_with_pipe!(UnixStream);
+use futures::stream::Stream;
+use futures::task::{Context, Poll};
```
vorner, line 120: Should these now come from `std`?

signal-hook-tokio/src/lib.rs
```diff
@@ -73,179 +114,37 @@ macro_rules! implement_signals_with_pipe {
...
-            let mut read_buf = ReadBuf::new(&mut buf);
-            match Pin::new(read).poll_read(ctx, &mut read_buf) {
-                Poll::Pending => Ok(false),
-                Poll::Ready(Ok(())) => Ok(true),
-                Poll::Ready(Err(error)) => Err(error),
-            }
-        }
-    }
+impl Stream for Signals {
+    type Item = c_int;
```
vorner, line 140: Seems I've missed this part when adding the support for extended info :-|. Well, fixing that doesn't belong into this PR, so let's ignore it for now.

vorner (commented):
Thank you! I somehow haven't managed to get around to this yet, so thanks for picking it up.

There are few details that need ironing out, but mostly this is what needed doing. Can you have a look at them?

signal-hook-tokio/src/lib.rs
bdash: Yup.

signal-hook-tokio/Cargo.toml
bdash: I updated both to use `~1`.

signal-hook-tokio/src/lib.rs
bdash: The lack of stable release for futures is an interesting point. Presumably the concern here is that the `Stream` trait changing in a future version. If that were to happen, or it were to be moved into the standard library as https://github.com/rust-lang/rfcs/pull/2996 proposes, it may be necessary to support multiple versions of `Stream` for a time. In the event that happens it seems reasonable to reintroduce feature flags to select which should be used. Do you see any need to do so proactively here?

 I don't see any direct dependency, direct or otherwise, on tokio-stream so I'm not sure clear what you're referring to there.

vorner: Well, the concern is not exactly about _changing_. I don't really expect that trait to have a different shape, there's not much space for other design that the single `poll_next` method. But even if the trait looks exactly the same, if it comes from a different crate (or different version), it becomes a different trait. And removing support for some older version would be a breaking change. That's why I would like to support the features proactively, so user can remove the dependency on 0.3 eventually, while we don't yet drop support for it. But as there would be only one instance right now, it would probably mean just slapping some `#[cfg(feature = "futures_0_3")] onto the `impl Stream for ...` block, to make sure the users that need it enable it.

I've looked at `tokio-stream` in more detail and it just reexports the `futures`' stream, so we are good there. Nevertheless, I've noticed that the `Stream` actually comes from `futures-core`. I'd suggest we depend on that one, as it is smaller dependency than `futures`.

bdash: Thanks for the explanation. I can see why it'd be useful to feature-gate it despite only a single version being relevant at present. I wasn't sure about the best way to achieve this. After some trial and error, I went with an approach that uses `cfg_if` to conditionally enable the `Stream` impl and supporting code while leaving the `Signals` type itself available unconditionally. Please let me know what you think.

I went ahead and switched the dependency on `futures` over to `futures-core` while leaving the references in the documentation as referring to `futures` for sake of familiarity for readers.

signal-hook-tokio/src/lib.rs
```diff
@@ -4,16 +4,57 @@
...
+//!     // Terminate the signal stream.
+//!     handle.close();
+//!     signals_task.await?;
+//!
+//!     Ok(())
+//! }
+//! ```
 
-#[cfg(any(feature = "support-v0_1", feature = "support-v0_3"))]
 macro_rules! implement_signals_with_pipe {
```
bdash, line 62: Is there any value in keeping this macro at this point, or should I go ahead and unwind it?

signal-hook-tokio/src/lib.rs
```diff
@@ -73,178 +117,42 @@ macro_rules! implement_signals_with_pipe {
...
-                Poll::Ok(Async::Ready(num_read)) => Ok(num_read > 0),
-                Poll::Err(error) => Err(error),
-            }
-        }
-    }
+use tokio::net::UnixStream;
+implement_signals_with_pipe!(UnixStream);
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "futures-v0_3")] {
```
vorner, line 124: Do you think it would be possible not to pull in the extra dependency? It seems our situation is not that complex to really need it. I was thinking about just putting the `#[cfg(...)]` on the `impl Stream for Signals` block. I'd probably say it's better to just suppress unused functions warnings instead with `#[cfg_attr(not(feature = "...", allow(dead_code))]`. I mean, the crate will be unusable without the feature anyway, so being clean at that point is not that important.

bdash: Conditionally suppressing the warnings is a great alternative that had not occurred to me. Thanks for the suggestion.

signal-hook-tokio/src/lib.rs
bdash: I'll look into it in a follow-up PR.

vorner:
Thank you :-). Looks good.

Do you want to follow up with that cleanup right away, or should I spin up a release?

PR #89, opened 2021-01-24
Remove the implement_signals_with_pipe macro

bdash:
It was previously being used to share logic that was common across multiple tokio versions. It's unnecessary now that
only one version is supported.


vorner (approved):

vorner:
Thank you. It's time for me to spin up a release.

PR #91, opened 2021-01-29
signal-hook-async-std: Switch to using async-io + futures-lite

zethra:
Switch to async-std's underlying async-io and futures-lite crates for
better compatibility with smol and using fewer dependencies in general.
Doing this also does not substantially changing the code, which is a
plus.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -701,7 +701,9 @@ dependencies = [
 name = "signal-hook-async-std"
 version = "0.2.0"
 dependencies = [
+ "async-io",
  "async-std",
+ "futures-lite",
  "libc",
  "serial_test",
  "signal-hook",
--- a/signal-hook-async-std/Cargo.toml
+++ b/signal-hook-async-std/Cargo.toml
@@ -21,7 +21,8 @@ maintenance = { status = "actively-developed" }
 
 [dependencies]
 libc = "~0.2"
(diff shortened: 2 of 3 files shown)


vorner (approved):
LGTM, but I have two questions:

* Is this a breaking change, or is the behaviour 100% the same from the outside? It seems this probably doesn't leak outside. But, does this work on every async-std version, or are only new versions based on top on this async-io?
* This can be plugged into async-std, but possibly into more things. So I wonder, if it should still be called `signal-hook-async-std` or something else.

zethra:
It shouldn't be a breaking change, however it may not work with every version of async-std, I'd have to test it. I think it will work, just with extra dependencies, but I'm not sure. As an alternative, we could maintain a separate crate with these changes.

zethra:
I don't think this is a breaking change but it probably should be tested against older versions of async-std to make sure it works. Alternatively, this patch could be a separate crate.

vorner:
Thinking about it again, let's not complicate things too much :-). It's still `0.*` crate without any explicit stability guarantees and if it works on the new version, it's probably fine ‒ if one updates one library, they should also update the others at the same time if possible anyway.

Adding another very similar crate would probably only confuse users and increase the maintenance burden. So let's keep the name as it is.

PR #101, opened 2021-04-02
Remove duplicate cfg_attr

iamihalcea:
This removes the duplicate `cfg_attr` that is leading to docs.rs build failures.

I ran the docs.rs build locally and it finished successfully - though strangely it's failing for 0.3.6 whereas their build finished successfully (even though both `cfg_attr`s are already there)


vorner:
Thanks for noticing and figuring out what it was. I've released the new version as 0.3.8 :crossed_fingers:

PR #105, opened 2021-05-07
Fix example code

feikesteenbergen:
The trailing closing delimiter is unexpected

### DIFF
--- a/signal-hook-tokio/src/lib.rs
+++ b/signal-hook-tokio/src/lib.rs
@@ -60,7 +60,7 @@
 //!
 //!     Ok(())
 //! }
-//! }
+//! # }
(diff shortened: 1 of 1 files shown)


vorner:
I kind of agree that the trailing comma is not aestetic, but I wouldn't call that one a „fix“.

The other one is actually needed. Try running `cargo test --all --all-features` (or see the CI, it should fail in a moment). It probably should be hidden with the `#`, though. Will you amend it, or should I take it over?

feikesteenbergen:
Amended the commit, the trailing comma is indeed aesthetic due to my compulsive `cargo fmt` usage nowadays!

vorner:
Thanks :-)

PR #106, opened 2021-05-20
Remove `SA_NOCLDSTOP` flag

VictorBulba:
Hey!

I am using ptrace with tokio and I need to wait for the SIGCHLD every time the process stops. But `SA_NOCLDSTOP` disables this feature. I don't know why it is set. You can find my investigation here https://github.com/tokio-rs/tokio/issues/3804 😄

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -691,12 +691,12 @@ dependencies = [
 
 [[package]]
 name = "signal-hook"
-version = "0.3.8"
+version = "0.4.0"
 dependencies = [
  "cc",
  "libc",
  "serial_test",
- "signal-hook-registry 1.3.0",
+ "signal-hook-registry 1.4.0",
 ]
 
 [[package]]
@@ -726,18 +726,18 @@ dependencies = [
 [[package]]
 name = "signal-hook-registry"
 version = "1.3.0"
-dependencies = [
- "libc",
- "signal-hook",
-]
-
-[[package]]
-name = "signal-hook-registry"
-version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "16f1d0fef1604ba8f7a073c7e701f213e056707210e9020af4528e0101ce11a6"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "signal-hook-registry"
+version = "1.4.0"
+dependencies = [
+ "libc",
+ "signal-hook",
 ]
 
 [[package]]
@@ -801,7 +801,7 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "pin-project-lite",
- "signal-hook-registry 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "signal-hook-registry 1.3.0",
  "tokio-macros",
  "winapi 0.3.9",
 ]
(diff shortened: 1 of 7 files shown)


vorner:
Hello

It is set because catching STOP is usually meaningful only for ptrace and similar things and they need to do _a lot_ of that ‒ and signal-hook (and tokio signaling) is probably not the best thing performance-wise for that. That was the reasoning back then.

I can understand why people might want, in some cases, want that, but most people don't, in practice. And I'm afraid this might be a breaking change for signal-hook-registry, which won't fly (bumping a major version to 2.0 would be a big task and this would be too small a motivation; even the bump of signal-hook itself to 0.4 seems like a big change, considering how many things depend on the 0.3 version and would have to migrate).

All in all, I think a proper solution would need some kind of configurability ‒ one hook asking for STOP ones too, while others continuing to be without it. I don't have a specific idea in mind right now, though.

VictorBulba:
Then I think separate feature for this behavior will solve the compatibility problem

vorner:
If you mean like a Cargo feature, then that doesn't seem like a good idea, because:

* Discoverability. It would be impossible to find (because it's quite deep in the dependency graph).
* That could still break unrelated code. The idea behind signal-hook is that multiple parts of application can independently register for a signal. If there are two libraries/parts of code depending on signal-hook-registry and one sets the feature flag, even the other one now starts getting „spurious“ SIGCHLD signals about stops that weren't there before.

So we either need a way for single _handler_ to ask for it, without bothering others, or deep thinking that actually this kind of spurious SIGCHLDs won't break other things (which I haven't done, I just suspect there's a change and I'm worried ‒ so if there's a reason why this can't break anything not already broken, then it would be fine ‒ but without the „big“ version bumps on signal-hook itself).

VictorBulba:
But they are not "spurious". They are real. And implicitly ignoring them is not a good solution.

VictorBulba:
I don't know if it will break the existing code or not. There are many ways to use signal handling. But it's strange if some code relies on this behavior. Because it is implicit and because the handling code usually has some specific checks. For example, when you are waiting for a process, you should check its status.

vorner:
So, I've been thinking about the possibilities and the effects on consumers. None is perfect, but in the end, merging this seems like the best one of them.

For the record, what I was thinking about:

# The risks

If someone assumes the current behaviour, they can expect that they get called because it terminated and if they eg. `waitpid`, they could get a deadlock or wait for too long or similar.

However, a library can't expect it to be the only one that forked something, so it might have as well been a different process that caused the SIGCHLD, so it needs to handle `waitpid` not being ready anyway.

An end-user process could assume that, but it's unlikely they'd handle processes manually.

So, the risks are somewhat low that this would break.

# The current problem

One doesn't get the stop/continue. Plus, it even stops anyone in the chain from getting them. This is bad. Furthermore, it seems signal-hook already assumes that these are delivered https://docs.rs/signal-hook/0.3.8/signal_hook/low_level/siginfo/enum.Chld.html#variant.Stopped. All in all, the longer I look at it, the more it looks like a bug.

# Alternatives

* Merge this. There's a low risk of breaking some application.
* Keep it as it is. This seems worse.
* Find some workaround, like introducing some kind of virtual SIG_CHLD_ALL, but that would be complicated to do (it would need C code introduced to signal-hook-registry and I've had to remove it from there once already), and arguably ugly API.

Anyway, thank you for discovering & fixing this, I'm going to merge and release some time soon.

VictorBulba:
Thanks! 😊

Issue #118, opened 2021-11-16
Linking fails on some Android versions


Hocuri:
When compiling on some Android versions, we get the following error:

```
[x86] SharedLibrary  : libnative-utils.so
jni/x86/libdeltachat.a(signal_hook-2530526319fba6eb.signal_hook.br2krarz-cgu.6.rcgu.o):signal_hook.br2krarz-cgu.6:function signal_hook::low_level::signal_details::emulate_default_handler::h0c9c85f8b16d90b7: error: undefined reference to 'sigemptyset'
jni/x86/libdeltachat.a(signal_hook-2530526319fba6eb.signal_hook.br2krarz-cgu.6.rcgu.o):signal_hook.br2krarz-cgu.6:function signal_hook::low_level::signal_details::emulate_default_handler::h0c9c85f8b16d90b7: error: undefined reference to 'sigaddset'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [obj/local/x86/libnative-utils.so] Error 1
```

See https://github.com/deltachat/deltachat-android/issues/2136. Summary:
- Looks like it is broken since vorner/signal-hook@b46473f
- For us, compilation is only failing when compiling for 32Bit, i.e. armeabi-v7a and x86, in debug mode. See the linked issue for thoughts on why it's failing specifically then.
- Probably you should exclude the code somehow for Android (possibly only exclude it for Android versions older than Android-5.0=API21 or only exclude it for 32bit-Android; both should already help in our issue), like you do for Windows

Thanks in advance!

vorner:
I'm probably OK with disabling it on the platforms that don't have it. But I don't have such system, so I'd ask you to:

* Track down which systems are impacted by this and which ones aren't (I do believe there are some androids this compiles for and I suspect they are even part of CI).
* Create a pull request. If you can also add such platform to the CI, it would be great.

The fact that the `libc` crate provides the function but it doesn't link seems to suggest a bug there (it probably should fail during compilation, not link time). Would you mind taking it to them too, please? (Disabling here still makes sense, either way it wouldn't be available).

Hocuri:
> Track down which systems are impacted by this and which ones aren't

Android systems below API 21 are impacted, see https://github.com/deltachat/deltachat-android/issues/2136.

However, there seems to be no way to do conditional compilation depending on the targeted Android version ([reference](https://doc.rust-lang.org/reference/conditional-compilation.html)].

[...]

PR #121, opened 2021-11-22
Don't use sigemptyset and sigaddset on Android

Hocuri:
Fix https://github.com/vorner/signal-hook/issues/118.

I successfully tested this by [overriding the dependency](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) to `signal-hook` in DeltaChat's local Cargo.toml:

```
[patch.crates-io]
signal-hook = { path = "../../../signal-hook" }
```

and running `./ndk-make.sh armeabi-v7a` and `./ndk-make.sh x86` (ndk-make.sh is our build script).

[64-bit support was added in android-21](https://stackoverflow.com/questions/49714180/why-android-ndk-standalone-toolchain-do-not-support-arm64-with-api-19-but-androi), so it's fine to only disable the code for 32-Bit Android code.

I didn't add this to the CI as this would be quite complicated because it's `ndk-build` that fails, not compiling (at least I couldn't come up with an easier way): We would have to let the CI download the Android SDK, write an Android.mk and Application.mk file, and run [ndk-build](https://developer.android.com/ndk/guides/ndk-build). And possibly do some other things to make `ndk-build` think that it's building an Android app.

### DIFF
--- a/src/low_level/signal_details.rs
+++ b/src/low_level/signal_details.rs
@@ -172,6 +172,16 @@ pub fn emulate_default_handler(signal: c_int) -> Result<(), Error> {
         DefaultKind::Stop => low_level::raise(SIGSTOP),
         DefaultKind::Term => {
             if let Ok(()) = restore_default(signal) {
+                // `sigemptyset` and `sigaddset` are not available on old Android versions before
+                // API 21 (Android 5.0).
+                // Unfortunately, we can't do conditional compilation based on the Android version,
+                // so we disable the code for all Android versions.
+                // As these old Android versions don't support 64-Bit-code anyway, we can leave
+                // the code enabled for 64-Bit-code.
+                #[cfg(not(all(
+                    target_os = "android",
+                    any(target_pointer_width = "32", target_pointer_width = "16")
+                )))]
                 #[cfg(not(windows))]
                 unsafe {
                     #[allow(deprecated)]


src/low_level/signal_details.rs
```diff
@@ -5,7 +5,7 @@ use std::mem;
 use std::ptr;
 
 use libc::{c_int, EINVAL};
-#[cfg(not(windows))]
+#[allow(unused_imports)]
 use libc::{sigset_t, SIG_UNBLOCK};
```
Hocuri, line 9: The other option here would be:
```
#[cfg(not(all(
    target_os = "android",
    any(target_pointer_width = "32", target_pointer_width = "16")
)))]
#[cfg(not(windows))]
use libc::{sigset_t, SIG_UNBLOCK};
```

src/low_level/signal_details.rs
```diff
@@ -172,6 +172,16 @@ pub fn emulate_default_handler(signal: c_int) -> Result<(), Error> {
...
             if let Ok(()) = restore_default(signal) {
+                // `sigemptyset` and `sigaddset` are not available on old Android versions before
+                // API 21 (Android 5.0).
+                // Unfortunately, we can't do conditional compilation based on the Android version,
+                // so we disable the code for all Android versions.
+                // As these old Android versions don't support 64-Bit-code anyway, we can leave
+                // the code enabled for 64-Bit-code.
+                #[cfg(not(all(
+                    target_os = "android",
+                    any(target_pointer_width = "32", target_pointer_width = "16")
```
Hocuri, line 183: I couldn't find out if 16-Bit Android code is a thing, which probably actually means that it's not. I included `target_pointer_width = "16"` here to be safe, but it's probably not needed.

vorner (commented):
Hello

I agree this could be used as kind of last-resort solution, so let's keep it here for now.

But I have to wonder if there's something more accurate. I've found the [autocfg](https://crates.io/crates/autocfg/) crate that discovers what the _compiler_ can do by running it with certain snippets. Now, that one doesn't actually _link_ the code, only _compile_, so it's not usable out of the box for this use case. Nevertheless, I think something in these lines might work:

* Create a sub-project (not necessarily as part of the workspace) which calls the sigemptyset function.
* Have a build.rs try compiling it (build scripts have access to bunch of env variables, so they know what cargo to call, etc).
* Based on that, emit a cfg directive.

Could you have a look how much work/code that would be?

Hocuri:
> * Have a build.rs try compiling it

But then we're still only compiling the code, not linking.

However, I just came up with yet another idea: Our build script (and therefore I assume other build scripts, too) sets the environment variable to e.g. `-D__ANDROID_API__=16`, so we could look for this environment variable in `build.rs` and then compile depending on it.

vorner:
> But then we're still only compiling the code, not linking.

I was thinking about a full build. Like, really calling something like `cd sigemptyset-check && cargo build --the-correct-build-params-targets-and-everything`. That would include linking and fail accordingly.

> DeltaChat's build script (and therefore I assume other build scripts, too)

That sounds somewhat backwards and fragile. Is it documented anywhere? Does everyone _have_ to set it?

When it comes to that, I'd probably prefer some kind of „workaround-for-broken-android“ feature or environment variable, it it can't be detected in a better way. It's still less preferable than being able to detect it in some „try-linking“ way.

Hocuri:
> something like `cd sigemptyset-check && cargo build --the-correct-build-params-targets-and-everything`. That would include linking and fail accordingly.

It's running `ndk-build`, a script from the Android SDK that fails, and whether it fails actually (also) depends on what's in `Application.mk`, an Android-specific config file that's in the parent directory. The `cargo build` command runs through, and it seems really unlikely that Cargo can be persuaded to do the job of `ndk-build`.

vorner:
So, the rust part is a library (`.a`)? I'm thinking of another way, providing our own „stub“ versions of these functions as weak symbols.

Does the android SDK promise which compiler will be used?

Hocuri:
> So, the rust part is a library (`.a`)?

Yes, `libdeltachat.a` (see the error message at https://github.com/vorner/signal-hook/issues/118).

> Does the android SDK promise which compiler will be used?

[The NDK uses the LLVM family of tools for building C/C++ code. These include Clang for compilation, LLD for linking, and other LLVM tools for other tasks](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#Introduction)

vorner:
I've tried to use the weak symbols to build something (hopefully) somewhat better. Please see #123. I'll be glad for some testing, as I don't have any android thing around.

vorner:
The alternative (hopefully better) version from #123 got merged instead. But thanks for this, it helped in developement of that branch.

PR #123, opened 2021-12-01
Fallback implementations of sigaddset & sigemptyset

vorner:
For certain android systems that are missing them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.3.11
+
+* Provide fallback sigaddset, sigemptyset on certain androids, as they are
+  missing them.
+
 # 0.3.10
 
 * Doc link fixes.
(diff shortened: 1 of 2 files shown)


Hocuri:
Unfortunately, this PR does not help. I get the same linking error as in #118 (I double-checked that if I switch to my `Hocuri:patch-1` branch from #121 instead of `android-hack`, it links fine again)

vorner:
@Hocuri Can you check this new version? It doesn't use the linker dark magic like the previous one, but at least tries to do the same thing as the missing functions.

Hocuri:
Yes, it works now, thanks a lot :tada::heart::heart::heart:!

vorner:
> Note though that I can't actually test anything, just see that it compiles and links fine.

That's better than me… so, let's hope this is working. If not, I guess there'll be another bug report soon :shrug:

Issue #119, opened 2021-11-16
Does not compile on Haiku due to usage of `SIGIO` constant


link2xt:
`signal-hook` attempts to import `SIGIO` here:
https://github.com/vorner/signal-hook/blob/49e45c068063282ac7ba23af79d18ef5e4e8b73b/src/lib.rs#L366
then uses it here:
https://github.com/vorner/signal-hook/blob/49e45c068063282ac7ba23af79d18ef5e4e8b73b/src/low_level/signal_details.rs#L50

`SIGIO` is a BSD name for the standard `SIGPOLL`, which is defined on Haiku:
https://github.com/haiku/haiku/blob/ae7d733d42435619cf76e0ab21a2c7af59dcf84f/headers/posix/signal.h#L142

Given that `SIGPOLL` is also obsolete according to the POSIX standard and is going to be removed in future versions, I think the correct solution is not to import either of them. `SIGIO`/`SIGPOLL` should not be generated unless you explicitly opt in via `fcntl`: https://www.gnu.org/software/libc/manual/html_node/Interrupt-Input.html

This is the only problem that prevents usage of `signal-hook` on Haiku, so fixing it will allow compiling the crate there.

vorner:
Hello

On systems that do have the signals, it is desirable to know how it behaves (the second link you have). So, my suggestion is:

* If it's available under a different name, provide an alias/define the constant (eg. `const SIGIO: c_int = SIGPOLL`)
* If it's not available at all, omit it from the list.

I don't really have much free time these days, so if you want to send a PR, it would speed things up. If you can also include the platform in CI, so we know it keeps working, it would be even  better.

PR #127, opened 2021-12-18
add haiku support

hoanga:
hello,

this adds haiku support per #119. if there is a more idiomatic way to handle this, would be open to suggestions.

the following shows a run of the unit tests  (after successful compilation) in which currently only one test doesn't pass.  my guess is that this test for haiku can probably be suppressed as it is unclear if datagram over unix sockets in haiku is currently available.

thanks!

```
> uname -a
Haiku shredder 1 hrev55602 Oct 29 2021 07:10:38 x86_64 x86_64 Haiku

> cargo test
  Downloaded proc-macro2 v1.0.33
  Downloaded 1 crate (41.0 KB) in 0.51s
   Compiling proc-macro2 v1.0.33
   Compiling syn v1.0.82
   Compiling smallvec v1.7.0
   Compiling instant v0.1.12
   Compiling parking_lot_core v0.8.5
   Compiling parking_lot v0.11.2
   Compiling quote v1.0.10
   Compiling serial_test_derive v0.5.1
   Compiling serial_test v0.5.1
   Compiling signal-hook v0.3.12 (/boot/home/src/git/rust-libs/signal-hook)
    Finished test [unoptimized + debuginfo] target(s) in 1m 09s
     Running unittests (target/debug/deps/signal_hook-dcf2f73007120844)

running 11 tests
test flag::tests::register_unregister ... ok
test low_level::channel::tests::multi_thread ... ok
test low_level::channel::tests::multiple ... ok
test low_level::channel::tests::new_empty ... ok
test low_level::channel::tests::overflow ... ok
test low_level::channel::tests::pass_value ... ok
test low_level::pipe::tests::register_dgram_socket ... FAILED
test low_level::pipe::tests::register_with_pipe ... ok
test low_level::pipe::tests::register_with_socket ... ok
test low_level::signal_details::test::existing ... ok
test low_level::signal_details::test::unknown ... ok

failures:

---- low_level::pipe::tests::register_dgram_socket stdout ----
Error: Os { code: -2147454955, kind: Uncategorized, message: "Address family not supported by protocol family" }
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /mnt/d/Rust/rust-x86/library/test/src/lib.rs:194:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -367,24 +367,21 @@ pub mod consts {
     /// namespace with other names. Also available in the [`consts`][crate::consts] directly (but
     /// with more constants around).
     pub mod signal {
-        #[cfg(not(any(windows, target_os = "haiku")))]
-        pub use libc::{
-            SIGABRT, SIGALRM, SIGBUS, SIGCHLD, SIGCONT, SIGFPE, SIGHUP, SIGILL, SIGINT, SIGIO,
-            SIGKILL, SIGPIPE, SIGPROF, SIGQUIT, SIGSEGV, SIGSTOP, SIGSYS, SIGTERM, SIGTRAP,
-            SIGTSTP, SIGTTIN, SIGTTOU, SIGURG, SIGUSR1, SIGUSR2, SIGVTALRM, SIGWINCH, SIGXCPU,
-            SIGXFSZ,
-        };
-
-        #[cfg(windows)]
-        pub use libc::{SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV, SIGTERM};
-
-        #[cfg(target_os = "haiku")]
+        #[cfg(not(windows))]
         pub use libc::{
             SIGABRT, SIGALRM, SIGBUS, SIGCHLD, SIGCONT, SIGFPE, SIGHUP, SIGILL, SIGINT,
             SIGKILL, SIGPIPE, SIGPROF, SIGQUIT, SIGSEGV, SIGSTOP, SIGSYS, SIGTERM, SIGTRAP,
             SIGTSTP, SIGTTIN, SIGTTOU, SIGURG, SIGUSR1, SIGUSR2, SIGVTALRM, SIGWINCH, SIGXCPU,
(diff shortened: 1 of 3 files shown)


vorner:
Hello

Have you tried the way with introducing an alias?

As for the pull request:
* Can the omission be done on single import item/one macro invocation instead? Something like: `pub use libc::{..., #[cfg(not(target_os = "haiku"))] SIGIO, ...`}? A completely different list is for windows because it's very different, but here it is similar. Or maybe pull it out, eg: `use libc::{most of the signals}; #[cfg(not(haiku))] use libc::SIGIO`?
* As for the failing test… I'm OK with skipping it, but it would be nice to verify that it's not really supported or check what's happening.

hoanga:
thanks for the comments. i am not sure how to approach the suggestion for alias without a more concrete example but can research it for the future.

i was not sure if the omission as you suggested would work but tried it out and looks like it works as hoped. thanks! i have pushed up a change that i hope looks a bit cleaner.

i did a check for datagram support for unix sockets and am rather sure that this is not a supported configuration in haiku so went ahead and disabled that test only for haiku.  see an excerpt below that i hope helps explain:

```
> cat main.c 

#include <stdio.h>
#include <stdlib.h>
#include <errno.h>
#include <string.h>
#include <sys/types.h>
#include <sys/socket.h>
#include <sys/un.h>

#define SOCK_PATH "socket_test_filename"

int main(void)
{
    int s, len;
    struct sockaddr_un local;
    char str[100];

    if ((s = socket(AF_UNIX, SOCK_DGRAM, 0)) == -1) {
        perror("socket");
        exit(1);
    }

    return 0;
}

> gcc -o sockettest main.c -lnetwork
> ./sockettest 
socket: Address family not supported by protocol family
```

vorner:
Thank you. I'm going to make a release soon.

PR #131, opened 2022-03-14
Add support for mio v0.8

pickfire:
Add support for mio-v0.8, not sure if this is the best way since it duplicates most of the v0.7, also not sure if there is a better way for this doing this, this allows helix which in terms depends on crossterm to be able to use signal-hook-mio with v0.8 to prevent crate duplication.

### DIFF
--- a/signal-hook-mio/tests/mio_0_7.rs
+++ b/signal-hook-mio/tests/mio_0_7.rs
@@ -1,15 +1,21 @@
-#![cfg(feature = "support-v0_7")]
+#![cfg(any(feature = "support-v0_7", feature = "support-v0_8"))]
 
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
 
+#[cfg(all(feature = "support-v0_7", not(feature = "support-v0_8")))]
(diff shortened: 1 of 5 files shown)


vorner:
Hello

I've somewhat busy time right now, I'll get around to it in a week or two. In general, having the support is welcome, I'll have a look at the details if this is the best way. I think some support for something already had a lot of duplication and was solving it by macros.

signal-hook-mio/tests/mio_0_7.rs
```diff
@@ -1,15 +1,21 @@
-#![cfg(feature = "support-v0_7")]
+#![cfg(any(feature = "support-v0_7", feature = "support-v0_8"))]
```
vorner, line 1: I think this'll not test both features if both are enabled. Can you create a separate one for mio 0.8? Can you also check that there's a CI setup with mio 0.8 enabled (there might be some with `--all-features` or some such).

vorner (changes requested):
Can you have a look at the tests?

vorner:
Thank you, released

pickfire:
I think you forgot to upload it to https://crates.io/crates/signal-hook-mio

vorner:
You're right. Fixed now.

PR #133, opened 2022-05-09
Add SIGINFO support on *BSD systems including macOS.

michael-yuji:
SIGINFO is a signal on *BSD systems to signal the process to output info related messages, for example progress in `dd` or stack trace in ruby. This PR add SIGINFO support by importing it from libc if available.

Tested platforms: FreeBSD and macOS.

### DIFF
--- a/src/low_level/signal_details.rs
+++ b/src/low_level/signal_details.rs
@@ -47,6 +47,10 @@ const DETAILS: &[Details] = &[
     s!(SIGHUP, Term),
     s!(SIGILL, Term),
     s!(SIGINT, Term),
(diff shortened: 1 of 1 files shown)


vorner:
Hello. Can you have a look and also extend the definitions in `src/low_level/signal_details.rs`?

Anyway, „support“ is a bit overstatement here. It's just convenience re-export. Whole signal-hook should work fine if you just pass the `libc::SIG*` constant or just the number.

michael-yuji:
@vorner Thanks for responding and the tip, just updated.

PR #136, opened 2022-09-28
Support AIX signal handler types

ecnelises:
The main differences for AIX:
- AIX doesn't have `MSG_DONTWAIT`, instead, use `MSG_NONBLOCK` (https://github.com/stephane/libmodbus/issues/294)
- AIX definition for sigaction handler is different: `sa_union { __su_handler: extern fn(c: ::c_int), __su_sigaction: extern fn(c: ::c_int, info: *mut siginfo_t, ptr: *mut ::c_void) }`

Note that AIX hasn't been an official Tier-3 Rust target. (https://github.com/rust-lang/compiler-team/issues/553)

### DIFF
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -159,12 +159,9 @@ impl Slot {
         // C data structure, expected to be zeroed out.
         let mut new: libc::sigaction = unsafe { mem::zeroed() };
         #[cfg(not(target_os = "aix"))]
-        let set_handler =
-            |action: &mut libc::sigaction, handler| action.sa_sigaction = handler as usize;
+        { new.sa_sigaction = handler as usize; }
         #[cfg(target_os = "aix")]
-        let set_handler =
-            |action: &mut libc::sigaction, handler| action.sa_union.__su_sigaction = handler;
-        set_handler(&mut new, handler);
+        { new.sa_union.__su_sigaction = handler; }
         // Android is broken and uses different int types than the rest (and different depending on
         // the pointer width). This converts the flags to the proper type no matter what it is on
         // the given platform.
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -158,7 +158,13 @@ impl Slot {
     fn new(signal: libc::c_int) -> Result<Self, Error> {
         // C data structure, expected to be zeroed out.
         let mut new: libc::sigaction = unsafe { mem::zeroed() };
-        new.sa_sigaction = handler as usize;
+        #[cfg(not(target_os = "aix"))]
+        let set_handler =
+            |action: &mut libc::sigaction, handler| action.sa_sigaction = handler as usize;
+        #[cfg(target_os = "aix")]
+        let set_handler =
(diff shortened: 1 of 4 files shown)


vorner (changes requested):
Hello

There are some style issues that would be nice to get addressed. Nevertheless, there's this bigger concern I have.

You mention that AIX is not yet a tier-3 Rust target. When trying to look at the libc crate, it doesn't seem to offer any aix target officially and grepping it for the word `aix` yields nothing, so is it even supported by libc? Or is it some kind of unofficial fork of the libc crate?

So I'd like to know how stable that target is, if it's likely to change (like, if the `libc` crate might paper-over the differences in the names, for example, in the future, which would break the currently proposed code). If it's not going to change, having it included in the CI in some way would be nice, because I have no other way to tell if it has any chance to work or if some future changes would break it.

ecnelises:
> You mention that AIX is not yet a tier-3 Rust target. When trying to look at the libc crate, it doesn't seem to offer any aix target officially and grepping it for the word aix yields nothing, so is it even supported by libc? Or is it some kind of unofficial fork of the libc crate?
>
> So I'd like to know how stable that target is, if it's likely to change (like, if the libc crate might paper-over the differences in the names, for example, in the future, which would break the currently proposed code). If it's not going to change, having it included in the CI in some way would be nice, because I have no other way to tell if it has any chance to work or if some future changes would break it.

AIX has not been an official Rust target yet. We have a proposal here https://github.com/rust-lang/compiler-team/issues/553 and multiple patches like https://github.com/rust-lang/libc/pull/2894. So this PR might need to be in draft status.

As for CI, we also plan to gradually set up CI for Rust toolchain after AIX target got accepted, but since it seems no public CI system supports AIX, we need to figure out how to support CI of these critical crates (also FYI @bzeq). Is that the necessity for now?

vorner:
> As for CI, we also plan to gradually set up CI for Rust toolchain after AIX target got accepted, but since it seems no public CI system supports AIX, we need to figure out how to support CI of these critical crates (also FYI @bzEq). Is that the necessity for now?

Not necessarily. Considering the state of the proposal, etc, I'd be OK merging it with the knowledge this is an experimental target, not guaranteed to work or compile despite the crate being past 1.0. That is, I'm OK accepting it and even publishing a new version if it helps you, but until there's some CI and such, I can't promise not to accidentally break it in some future version :shrug:. If that's fine for you, we can go forward with this.

ecnelises:
> Not necessarily. Considering the state of the proposal, etc, I'd be OK merging it with the knowledge this is an experimental target, not guaranteed to work or compile despite the crate being past 1.0. That is, I'm OK accepting it and even publishing a new version if it helps you, but until there's some CI and such, I can't promise not to accidentally break it in some future version 🤷. If that's fine for you, we can go forward with this.

Thanks. Now `powerpc64-ibm-aix` has been accepted as Tier-3 target. I'm okay with potential build break from future change due to lack of CI.

vorner (approved):

vorner:
Thank you, released…

ecnelises:
Thanks!

PR #152, opened 2023-12-02
doc: fixed typo

cppcoffee:


vorner:
This is not a typo. These are typographic quotes that differ between opening and closing ones ‒ which most people don't use out of laziness because they are harder to type, similar to people not making the difference between short dash, long dash and a minus sign.

I'm open for discussion about if the opening quotes should be at the top or the bottom of the line, though 😇

PR #156, opened 2024-03-27
Add wait_for_shutdown_signal convenience function.

mleonhard:
Hi vorner,
I started adding a `wait_for_shutdown_signal` function to https://crates.io/crates/permit .  Then I thought that it would probably be cleaner to put it into signal-hook.  What do you think?

This PR also deletes Cargo.lock to make CI use the latest versions of dependencies.

Cheers,
mleonhard


vorner:
First of, I prefer to keep the Cargo.lock in the repo. I prefer to update manually. Don't worry, the downstream crates don't consider Cargo.lock from libraries, it's only from development ‒ I've had some trouble with the CI updating on its own.

As for the pull request. The general idea seems fine, with some ideas for improvement:

* I wonder why just _one_ of the crates has a feature for it and not the others.
* Generally, the close is not necessary. It is implicit when the instance is dropped.
* I don't think it's a good idea to rely on CI running fast enough, therefore checking some specific times.
* I'd suggest putting the list of signals into a constant somewhere, so it's shared between all the crates.
* I'm not sure if the base case (blocking) version is particularly useful.

vorner:
>  I don't understand why the crate uses futures and aliases the dependencies.

If I remember correctly (it's a long time ago), it was in times when tokio didn't have any streaming support yet. I suspect some kind of cleanup and moving to using that might be a good idea some day.

> I expected this, but the docs didn't say it and the code does not impl Drop. Also the examples say to close:

It's on the DeliveryState in src/iterator/backend.rs.

The `close` is so that, when one thread is blocked on the iterator, another thread can remotely terminate the iterator and unblock it. It is used in the examples in such a way.

> The existing tests use sleeps:

I'm not against using `sleep` in the tests. However, in the past, I had some bad experience on relying on a test running _fast enough_ ‒ because the CI machine was severely overloaded. It was generating false failures, because something didn't happen in time. If you look at the existing sleep tests, it checks that something _doesn't_ happen. On a really slow CI, it could cause a false pass, not false failure, which is AFAIK better.

> Would you please share your motivations for this change?

Well, one is that I don't like duplicated code.

But as you suggest, it is a commonly used list of signals and a user might find it useful for other purposes. So putting it into the base `signal-hook` crate, it is available both to our own use but also for a user.

> Would you please elaborate on this?

Well, the primitive contains a lot of code to make sure it can make a distinction between different signals. It's complicated because of that and relies on a pipe(like) _combined_ with an array of atomic flags. The application then can consume the signals and dispatch different actions based on what comes, possibly including extended info.

The `wait_for_shutdown_signal` kind of throws all that out of the window and just… uses the same action for all. We need only a _wakeup_ (unblocking). Which can be done either by a semaphore (sem_post _is_ allowed in a signal handler) or by a pipe(like) primitive alone. The crate provides the pipe primitive ‒ so what you can do is register the same write end to multiple signals and then wait in a blocking wait to read a byte out of it. Now looking at what the crate offers, it seems not to offer a way to directly register the same write end multiple times ‒ so it would have to be duped in some way or provide an abstraction for it.

PR #158, opened 2024-04-12
QNX Neutrino support

SebastianSchildt:
This adds support for QNX Neutrino operating system

This would help us with https://github.com/tokio-rs/tokio/pull/6421 and https://github.com/eclipse/kuksa.val/issues/741 so merging this would be appreciated

@AkhilTThomas fyi


vorner:
I guess such a little change can be added, even if we don't have a CI or anything for that platform. Released.

PR #161, opened 2024-07-21
mio option update proposal, for having haiku support mostly.

devnexen:


vorner:
Hello

I'm not opposed to adding support for the newest version and I'm happy that you send it as a pull request. However, I don't think this way works ‒ if I'm not overlooking it, you're missing the relevant parts inside the code and in the tests. Look for eg. `[cfg(feature = "support-v0_9")]` how it is done.

vorner (approved):

vorner:
OK, released. Thank you :+1:

PR #166, opened 2024-11-13
expose `si_status` for SIGCHLD

### LINKED ISSUE #165: Expose `siginfo_t` `si_status` field as `Option<c_int>` in `siginfo::Process` on SIGCHLD
It's the missing piece for full `SIGCHLD` handling using `signal-hook`. We already have `Origin::Chld` for detecting child exit, child death, child stop, etc.

The way `libc` exposes `siginfo_t` is a bit weird, but I see that `signal-hook` has a bit of C code that handles the extraction, presumably so as to avoid `libc`s weirdness, so I think it would be reasonably straightforward.

I could make a PR for this if desired.

LunarLambda:
Problem: there is no way to get the exit status or signal of a child process when handling SIGCHLD.

Solution: expose `siginfo_t`'s `si_status` field in `Process` when extracting info for SIGCHLD.

Let me know if I should drop or change the version bump commit. It might not be entirely correct.

closes #165

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,17 +1012,17 @@ dependencies = [
 
 [[package]]
 name = "signal-hook"
-version = "0.3.17"
+version = "0.4.0"
 dependencies = [
  "cc",
  "libc",
  "serial_test 0.7.0",
- "signal-hook-registry 1.4.2",
+ "signal-hook-registry 1.4.3",
 ]
 
 [[package]]
 name = "signal-hook-async-std"
-version = "0.2.2"
+version = "0.2.3"
 dependencies = [
  "async-io 1.13.0",
  "async-std",
@@ -1034,7 +1034,7 @@ dependencies = [
 
 [[package]]
 name = "signal-hook-mio"
-version = "0.2.4"
+version = "0.2.5"
 dependencies = [
  "libc",
  "mio 0.6.23",
@@ -1057,7 +1057,7 @@ dependencies = [
 
 [[package]]
 name = "signal-hook-registry"
-version = "1.4.2"
+version = "1.4.3"
 dependencies = [
  "libc",
  "signal-hook",
@@ -1065,7 +1065,7 @@ dependencies = [
 
 [[package]]
 name = "signal-hook-tokio"
-version = "0.3.1"
+version = "0.3.2"
 dependencies = [
  "futures",
  "futures-core",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "signal-hook"
-version = "0.3.17"
+version = "0.4.0"
 authors = [
     "Michal 'vorner' Vaner <vorner@vorner.cz>",
     "Thomas Himmelstoss <thimm@posteo.de>",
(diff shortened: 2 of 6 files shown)


vorner:
Apparently, the CI degraded over time, so it's failing a lot :-(.

How portable is this field? Is it supported on every weird platform that exists?

Also, why is it needed? When you receive a SIGCHLD, aren't you supposed to waitpid on it anyway?

As for the bump of version, it also needs a note in the changelog. And I don't think there's a need to go from 0.3 to 0.4.

LunarLambda:
si_status is part of POSIX. However, depending on platform it might be stuffed inside of a union. Hence the SIGCHLD check in extract.c

Yes, you are generally meant to wait on the process to reap then, however this allows getting all the info in one place, as well as std::process::Child only allows you to wait for exit, so you can't handle things like a child processing suspending or stopping due to job control (SIGSTOP, SIGTOUT) or similar.

I can remove the version bump and add release notes. I thought it would be necessary since it adds a new public field to Process.

LunarLambda:
I had missed the struct being non_exhaustive, that's good then.

I can take a look at the CI, but I don't have a ton of experience with that. I did run `cargo test` over the whole workspace at least.

blackbit42:
Bump!
This would be a very useful feature.
Besides - and please let me know if you think this is too far off-topic for this PR - it would be helpful to expose all or at least some additional fields of `siginfo_t`.
@LunarLambda would you be in the mood to add that?

LunarLambda:
> Apparently, there wasn't much activity 😇. The CI got revived a bit recently, so maybe we just need to rebase it. Should I take it over?
> 
> As for „all“ the fields of siginfo_t. The problem is, platforms have different ideas of what's inside it and what exactly it means. I'm not against adding more fields in general, as long as it is done in some platform independent way (either fields that are available „everywhere“, according to CI, or that they are available conditionally only on the platforms that have them). Not sure if it belongs to this particular PR, though… it might need some iteration with the CI to check.

I'm currently at a convention. I could take a look at rebasing the PR in ~32 hours or so from now. Still don't really have anything to say regarding the CI ^^. Still interested in getting this pushed through though.

vorner:
> I'm currently at a convention. I could take a look at rebasing the PR in ~32 hours or so from now. Still don't really have anything to say regarding the CI ^^.

Please do so at a time of your convenience :-).

As for the CI - as I mentioned, it got revived in current master. That's why I'm suggesting to rebase it there.

blackbit42:
Huh, what happened?

LunarLambda:
I've archived, deleted, or moved most of my GitHub repositories today as I will no longer be using the platform for anything except submitting patches. Unfortunately, I checked my fork and it was a very very large number of commits behind. I don't really have the time or energy to rebase this right now, especially with the catastrophic state of GitHub's CI (as seen again by the outage today). I also don't currently need this functionality for any of my projects.

The relevant part of the patch is still available under this URL, should you wish to adapt it: https://github.com/vorner/signal-hook/pull/166/changes/6dcb5953c2ae3fe7fe5ae87bc6f469f853f54c99

I apologise for any inconvenience.

blackbit42:
Thank you, godspeed.

PR #168, opened 2025-01-29
Update LICENSE-MIT

maximevtush:
Updated the copyright year from 2017 to 2025 in the LICENSE file

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2017 tokio-jsonrpc developers
+Copyright (c) 2025 tokio-jsonrpc developers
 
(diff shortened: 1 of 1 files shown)


vorner:
Hmm :thinking: 

For one, I'm not sure it _should_ be updated to 2025 if there was no change made in 2025.

But apparently, when looking at it, even the `tokio-jsonrpc developers` part is just a copy-paste from other project and obviously wrong.

I do wonder, how important is it to even have the file in the repository / or the first line? Isn't it enough that git preserves all the copyright info in the commits and that we say what the license is in the Cargo.toml?

vorner:
Deleting the year and fixing the owner might be the best :-)

PR #171, opened 2025-04-09
minor: Close codeblock in `low_lewel::pipe` docs

the-mikedavis:
:wave:

Super minor docs fix here - the codefence for the `pipe` module doc wasn't closed. Rustdoc doesn't seem to care but it can mess up some editors' syntax highlighting. Thanks!


vorner:
Thank you

PR #175, opened 2025-04-20
Update async-io and futures-lite to v2

### LINKED ISSUE #172: signal-hook-async-std is using outdated dependancies
signal-hook-async-std is use 1.x versions of async-io and futures-lite instead of the up to date 2.x versions.

gaykitty:
Fixes #172.


vorner:
Thank you, released.

PR #179, opened 2025-07-24
Clarify safety requirements for register.

### LINKED ISSUE #178: `signal_hook::low_level::register` should document that access to thread local variables is unsafe from signal handlers
I think this program exhibits undefined behavior, without violating any of the [safety requirements documented for `signal_hook::low_level::register`](https://docs.rs/signal-hook/0.3.18/signal_hook/low_level/fn.register.html#safety):

```rust
use std::cell::RefCell;
use std::sync::Arc;
use std::sync::atomic::{AtomicUsize, Ordering};
use std::thread::{sleep, spawn};
use std::thread_local;
use std::time::Duration;

use signal_hook::consts::SIGQUIT;
use signal_hook::low_level::register;

thread_local! {
    static CELL: RefCell<usize> = const { RefCell::new(0) };
}

fn main() {
    unsafe { register(SIGQUIT, handler) }.unwrap();
    let output = Arc::new(AtomicUsize::new(0));
    spawn({ // Spawning this thread isn't strictly required to trigger the UB, but it helps to observe run's view of `CELL` without polluting the disassembly of run with function calls
        let output = output.clone();
        move || loop {
            println!("x = {}", output.load(Ordering::Relaxed));
            sleep(Duration::from_secs(1));
        }
    });
    run(&output);
}

#[inline(never)]
fn run(output: &AtomicUsize) -> ! {
    loop {
        CELL.with_borrow_mut(|x| {
            output.store(*x, Ordering::Relaxed);
            *x = x.wrapping_add(1);
        });
    }
}

#[inline(never)]
fn handler() {
    CELL.with_borrow_mut(|x| {
        *x = 42;
    });
}
```

Specifically, the behavior I am currently observing on release builds with rustc 1.88.0 is that sending SIGQUIT to the process never actually resets the counter in `CELL`.

This happens, because both `run` and `handler` are optimized under the assumption that during their runtime, no other unrelated code can run and modify `CELL`. This assumption is of course violated, since unbeknownst to the compiler, a SIGQUIT signal can actually interrupt `run` and jump into `handler` instead.

The assembly produced for this program looks like `handler` checks if `CELL` is borrowed, but does not actually update the borrowing state in memory. `run` is optimized even more extremely. Besides an initial check of whether it is borrowed, `CELL` is never actually accessed and `*output` is instead just incremented directly.

This problem is specific to thread local statics, because regular statics are required to be `Sync` and so is the closure accepted by `register`. Thread locals are a special case here, because they don't need to be `Sync`, but their use inside a signal handler cannot be detected from the type of the closure.

I think this restriction should be documented in the safety section of the function. Maybe something along the lines of:

> Thread local memory may not be accessed from within the signal handler, unless the access is properly synchronized with other accesses of the thread local in the thread through use of a [`compiler_fence`](https://doc.rust-lang.org/stable/core/sync/atomic/fn.compiler_fence.html) or stronger. Notably, common providers of unsynchronized interior mutability such as `Cell` or `RefCell` are not suitable for sharing thread locals between signal handlers and the thread without additional synchronization.

vorner:
Closes #178.

### DIFF
--- a/signal-hook-registry/src/lib.rs
+++ b/signal-hook-registry/src/lib.rs
@@ -462,18 +462,34 @@ const FORBIDDEN_IMPL: &[c_int] = &[SIGKILL, SIGSTOP, SIGILL, SIGFPE, SIGSEGV];
 ///
 /// # Safety
 ///
-/// This function is unsafe, because the `action` is run inside a signal handler. The set of
-/// functions allowed to be called from within is very limited (they are called async-signal-safe
-/// functions by POSIX). These specifically do *not* contain mutexes and memory
-/// allocation/deallocation. They *do* contain routines to terminate the program, to further
-/// manipulate signals (by the low-level functions, not by this library) and to read and write file
-/// descriptors. Calling program's own functions consisting only of these is OK, as is manipulating
-/// program's variables ‒ however, as the action can be called on any thread that does not have the
-/// given signal masked (by default no signal is masked on any thread), and mutexes are a no-go,
-/// this is harder than it looks like at first.
+/// This function is unsafe, because the `action` is run inside a signal handler. While Rust is
+/// somewhat vague about the consequences of such, it is reasonably to assume that similar
+/// restrictions as specified in C or C++ apply.
+///
+/// In particular:
+///
+/// * Calling any OS functions that are not async-signal-safe as specified as POSIX is not allowed.
+/// * Accessing globals or thread-locals without synchronization is not allowed (however, mutexes
+///   are not within the async-signal-safe functions, therefore the synchronization is limited to
+///   using atomics).
+///
+/// The underlying reason is, signals are asynchronous (they can happen at arbitrary time) and are
+/// run in context of arbitrary thread (with some limited control of at which thread they can run).
+/// As a consequence, things like mutexes are prone to deadlocks, memory allocators can likely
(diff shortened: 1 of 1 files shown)


signal-hook-registry/src/lib.rs
```diff
@@ -462,18 +462,31 @@ const FORBIDDEN_IMPL: &[c_int] = &[SIGKILL, SIGSTOP, SIGILL, SIGFPE, SIGSEGV];
...
-/// functions by POSIX). These specifically do *not* contain mutexes and memory
-/// allocation/deallocation. They *do* contain routines to terminate the program, to further
-/// manipulate signals (by the low-level functions, not by this library) and to read and write file
-/// descriptors. Calling program's own functions consisting only of these is OK, as is manipulating
-/// program's variables ‒ however, as the action can be called on any thread that does not have the
-/// given signal masked (by default no signal is masked on any thread), and mutexes are a no-go,
-/// this is harder than it looks like at first.
-///
-/// As panicking from within a signal handler would be a panic across FFI boundary (which is
-/// undefined behavior), the passed handler must not panic.
```
Kritzefitz, line 476: This info about panics still seem relevant, but isn't reflected in the new text. Should it still be there?

Kritzefitz (commented):
Apart from the question above, this looks good to me.

PR #188, opened 2025-10-25
Allow mio 1.1+

dbanty:
With mio being 1.x it's now, in theory at least, stable. So anything less than a major version bump should be compatible.

This PR allows any mio 1.x (right now 1.1) so downstream consumers can keep their dependencies up to date easier.


vorner:
Thank you, I've released a new version with that.

PR #193, opened 2025-11-23
Various CI fixes

hanna-kruppe:
Took a bunch of trial and error I mostly managed to preserve the existing test coverage, as I understood it. Notable exceptions:

- Ancient Rust versions no longer run tests on macOS runners because the ARM64 ones aren't a supported target on ancient Rust and the Intel macOS runners that still exist ran into some weird linker error I didn't know how to resolve.
- Some of the "weird targets" have been removed entirely (emscripten) or no longer have pre-built artifacts (mips-unknown-linux-musl). I haven't checked if there's another MIPS target that could be substituted, but I added s390x Linux so there's at least one big endian target in the mix.
- I removed the Solaris build because [mio doesn't compile on it any more](https://github.com/hanna-kruppe/signal-hook/actions/runs/19614258647/job/56164552544).

### DIFF
--- a/.github/actions-rs/grcov.yml
+++ b/.github/actions-rs/grcov.yml
@@ -1,6 +1,0 @@
-branch: true
-ignore-not-existing: true
-llvm: true
-filter: covered
-output-type: lcov
-output-path: ./lcov.info


hanna-kruppe:
I didn't find a nice, maintainable way to have CI omit only the extended-siginfo feature only when testing on Windows. So instead I've gated the *functionality* of the feature behind `cfg(not(windows))`, mirroring `low_level::pipe`. In other words, the feature can be enabled on Windows but it won't do anything, other than enabling an unused `cc` dependency.

After some more whack-a-mole with windows-specific warnings and errors, the only remaining failure is the coverage workflow, that seems to be due to repository configuration (it worked on my fork), so I can't resolve that.

PR #198, opened 2025-12-27
Remove `Cargo.lock.old`

cakebaker:
I think `Cargo.lock.old` was committed by mistake and so this PR removes the file.


vorner:
No, it was committed on purpose. This variant fixes some older version of dependencies so they still build with old versions of rustc in the CI: https://github.com/vorner/signal-hook/blob/master/ci-check.sh#L24-L26

PR #209, opened 2026-08-03
low_level::pipe: set close-on-exec via F_SETFD, not F_SETFL

### LINKED ISSUE #206: `low_level::pipe`: `set_flags` documents close-on-exec but passes `O_CLOEXEC` to `F_SETFL`
## Summary

`WakeFd::set_flags` in `src/low_level/pipe.rs` is documented as

```rust
/// Sets close on exec and nonblock on the inner file descriptor.
fn set_flags(&self) -> Result<(), Error> {
```

but it sets **only** non-blocking. The close-on-exec half is a no-op, because `O_CLOEXEC` is
passed to `F_SETFL`, which manipulates file *status* flags, whereas close-on-exec is a file
*descriptor* flag set through `F_SETFD`/`FD_CLOEXEC`.

The consequence is that a self-pipe write end registered through the public
`low_level::pipe::register` / `register_raw` **survives `exec`** and is inherited by child
processes, contrary to the documented behaviour.

## The code

`src/low_level/pipe.rs:102-115` (0.3.18):

```rust
    /// Sets close on exec and nonblock on the inner file descriptor.
    fn set_flags(&self) -> Result<(), Error> {
        unsafe {
            let flags = libc::fcntl(self.as_raw_fd(), libc::F_GETFL, 0);
            if flags == -1 {
                return Err(Error::last_os_error());
            }
            let flags = flags | libc::O_NONBLOCK | libc::O_CLOEXEC;
            if libc::fcntl(self.as_raw_fd(), libc::F_SETFL, flags) == -1 {
                return Err(Error::last_os_error());
            }
        }
        Ok(())
    }
```

`F_GETFL` / `F_SETFL` read and write the **file status flags**. POSIX specifies that `F_SETFL`
acts only on `O_APPEND`, `O_DSYNC`, `O_NONBLOCK`, `O_RSYNC` and `O_SYNC`; Linux's `fcntl(2)`
states the same and adds that other bits are ignored. `O_CLOEXEC` is not among them, so the
bit is silently discarded.

Close-on-exec lives in a different namespace entirely, reached with `F_GETFD`/`F_SETFD` and the
constant `FD_CLOEXEC`. The two constants are not even the same value: `FD_CLOEXEC` is `0x1`,
while `O_CLOEXEC` on `x86_64-unknown-linux-gnu` is `0o2000000`. So the OR does not accidentally
set the right bit in the wrong namespace either.

## Second source: the standard library draws exactly this distinction

`library/std/src/sys/fd/unix.rs` (rustc 1.95.0) keeps the two apart deliberately:

```rust
    pub fn set_cloexec(&self) -> io::Result<()> {
        unsafe {
            let previous = cvt(libc::fcntl(self.as_raw_fd(), libc::F_GETFD))?;
            let new = previous | libc::FD_CLOEXEC;
            if new != previous {
                cvt(libc::fcntl(self.as_raw_fd(), libc::F_SETFD, new))?;
            }
            Ok(())
        }
    }
```

versus, for the status-flag case:

```rust
    pub fn set_nonblocking(&self, nonblocking: bool) -> io::Result<()> {
        unsafe {
            let previous = cvt(libc::fcntl(self.as_raw_fd(), libc::F_GETFL))?;
            let new = if nonblocking { previous | libc::O_NONBLOCK } else { previous & !libc::O_NONBLOCK };
            if new != previous {
                cvt(libc::fcntl(self.as_raw_fd(), libc::F_SETFL, new))?;
            }
            Ok(())
        }
    }
```

`set_flags` merges these two operations into one `F_SETFL` call, which is correct for the
`O_NONBLOCK` half and inert for the `O_CLOEXEC` half.

## Third source: mio does it correctly, seven times, in the same dependency graph

The standard library's practice could be dismissed as one project's style. It is not one
project's style. **mio 1.2.1 — which sits beside signal-hook in the same crossterm dependency
closure, targets the same platforms, and solves the same problem — never once mixes the two
namespaces, across seven independent sites.** Line numbers are from mio 1.2.1, and each was
read at site during that crate's `cargo vet` certification (bead `warden-sxcz`).

| # | Site | What it does |
| --- | --- | --- |
| 1 | `src/sys/unix/selector/epoll.rs:29` | `epoll_create1(EPOLL_CLOEXEC)` — atomic at creation |
| 2 | `src/sys/unix/waker/eventfd.rs` | `eventfd(EFD_CLOEXEC \| EFD_NONBLOCK)` — atomic at creation |
| 3 | `src/sys/unix/net.rs:29` | `socket_type \| SOCK_NONBLOCK \| SOCK_CLOEXEC` — atomic in the `socket()` call |
| 4 | `src/sys/unix/net.rs:70-80` | the fallback: `fcntl(F_SETFL, O_NONBLOCK)` **and** `fcntl(F_SETFD, FD_CLOEXEC)` as separate calls |
| 5 | `src/sys/unix/tcp.rs:79-84`, `:109-134` | `accept4(2)` with `SOCK_CLOEXEC\|SOCK_NONBLOCK` where available; the fallback uses `F_SETFD` (`:117`) and `F_SETFL` (`:128`) separately |
| 6 | `src/sys/unix/pipe.rs:8-64` | `pipe2(fds, O_CLOEXEC \| O_NONBLOCK)` on Linux; the fallback does `F_SETFL` and `F_SETFD` per descriptor as separate calls (`:51-52`) |
| 7 | `src/sys/unix/pipe.rs:589-619` | `set_nonblocking`: `ioctl(FIONBIO)` on most platforms, and on aix/illumos/solaris/vita an `F_GETFL`/`F_SETFL` pair setting `O_NONBLOCK` **and only** `O_NONBLOCK` |

Two of these are worth quoting directly, because they bracket the defect.

**Site 6 is the closest analogue to `set_flags` itself** — the same two flags, on the same
object type (a pipe), in the same kind of no-`pipe2` fallback:

```rust
    // mio 1.2.1, src/sys/unix/pipe.rs, fallback path
    syscall!(fcntl(fd, libc::F_SETFL, libc::O_NONBLOCK))?;
    syscall!(fcntl(fd, libc::F_SETFD, libc::FD_CLOEXEC))?;
```

Two calls, two commands, two namespaces. `set_flags` collapses exactly this pair into one
`F_SETFL` and loses the close-on-exec.

**Site 7 is the purest**, showing what `F_SETFL` looks like when it is used correctly — for a
file *status* flag, alone, with the other flags preserved and the write skipped if nothing
changed. Nowhere in mio does `F_SETFL` carry `O_CLOEXEC`.

Site 3's wasi arm (`net.rs:32-33`) omits `SOCK_CLOEXEC` with a comment explaining that wasi has
no `fork`/`exec` so the flag "neither exists nor is relevant" — a reasoned omission rather than
an oversight, which is the contrast that makes the seven-for-seven record meaningful.

So the claim here does not rest on the man page plus one second opinion. **Two independent
crates in signal-hook's own neighbourhood keep `F_SETFD` and `F_SETFL` apart wherever the
question arises — mio at all seven sites above, the standard library in its
`set_cloexec`/`set_nonblocking` pair — and `set_flags` is the single place in warden's entire
dependency graph where the two are merged.**

## Reachability

`set_flags` has exactly one call site: `pipe.rs:190`, inside `register_raw`, and only on the
`WakeMethod::Write` branch — i.e. only when the supplied descriptor is **not** send-capable
(a real pipe rather than a socket). The `WakeMethod::Send` branch returns without calling it,
relying on `MSG_DONTWAIT` instead.

Within signal-hook itself, `register_raw`'s callers are `pipe::register` (`pipe.rs:210`), the
module doc-example (`pipe.rs:68`), and one unit test (`pipe.rs:259`). The iterator machinery
does not use this path: `iterator/backend.rs:44` calls `pipe::wake(fd, WakeMethod::Send)`
directly, and `iterator/mod.rs:151` builds its pair with `UnixStream::pair()`.

**There are real external callers, and the most prominent one is crossterm.**
`crossterm-0.28.1/src/event/source/unix/tty.rs:72` calls
`pipe::register(libc::SIGWINCH, sender)?` to receive terminal-resize notifications.

That call does **not** hit the bug, and the reason is worth stating precisely because it is not
"the function is never called". `sender` comes from crossterm's `nonblocking_unix_pair()`
(`tty.rs:51-56`), i.e. `UnixStream::pair()`. `register_raw` probes the descriptor with a
zero-byte `libc::send(pipe, &[] as *const _, 0, MSG_NOWAIT)` (`pipe.rs:179`); for a socket this
succeeds, so the `WakeMethod::Send` arm is taken (`pipe.rs:181-184`) and `set_flags` is never
reached. Close-on-exec is nevertheless set on that descriptor — by the standard library, which
creates socket pairs with `SOCK_CLOEXEC` — and crossterm sets non-blocking itself (`tty.rs:54`).

So the defect bites exactly the callers who hand a **pipe** (or any non-send-capable
descriptor) to `low_level::pipe::register` / `register_raw` and rely on the documented
close-on-exec. Socket callers are saved by their descriptor's provenance rather than by
anything in this function.

## Impact

Low-to-moderate, and mostly a documentation-versus-behaviour hazard:

- A program that registers a self-pipe with these functions and then spawns a child leaks the
  write end into the child. The child holds the pipe open, so the parent's read end never sees
  EOF on child exit, and the descriptor is visible to unrelated code in the child.
- No memory-unsafety, no UB. The `unsafe` block is sound; only the effect is missing.
- Callers who pass sockets (the documented recommendation, and what the crate's own iterator
  uses) are unaffected, because that path never calls `set_flags`.

## Suggested fix

Split the two, matching the standard library:

```rust
    fn set_flags(&self) -> Result<(), Error> {
        unsafe {
            let flags = libc::fcntl(self.as_raw_fd(), libc::F_GETFL, 0);
            if flags == -1 {
                return Err(Error::last_os_error());
            }
            if libc::fcntl(self.as_raw_fd(), libc::F_SETFL, flags | libc::O_NONBLOCK) == -1 {
                return Err(Error::last_os_error());
            }
            let fd_flags = libc::fcntl(self.as_raw_fd(), libc::F_GETFD);
            if fd_flags == -1 {
                return Err(Error::last_os_error());
            }
            if libc::fcntl(self.as_raw_fd(), libc::F_SETFD, fd_flags | libc::FD_CLOEXEC) == -1 {
                return Err(Error::last_os_error());
            }
        }
        Ok(())
    }
```

Alternatively, if inheriting the descriptor is considered acceptable, amend the doc comment so
it stops promising close-on-exec.

## How this was found

Line-by-line read of signal-hook 0.3.18 during a `cargo vet` certification of warden's
dependency graph. The mismatch was noticed by reading the doc comment against the `fcntl`
command it uses; the reading was then confirmed against the standard library's own
`set_cloexec`/`set_nonblocking` pair rather than from memory of the man page. Reachability was
traced mechanically over every call site of `set_flags`, `register_raw` and `WakeMethod`.

The seven mio sites were added later, during the line-by-line certification of mio 1.2.1 (bead
`warden-sxcz`), which read the whole compiled-reachable ring of that crate on the same target.
They were not sought out to support this report — they were encountered one at a time while
reading for other reasons, which is why they are spread across five files and three subsystems.

Not verified by execution: the audit host is Windows, so no observation of the descriptor bit
after `F_SETFL` was made on a live unix process. The claim rests on the documented semantics of
`F_SETFL` plus the standard library's contrasting practice.

## Upstream context, and a correction to the reasoning that closed #41

Checked against the default branch before filing: `src/low_level/pipe.rs:103-112` is unchanged.
The doc comment still reads "Sets close on exec and nonblock on the inner file descriptor", and
the body still ORs `O_CLOEXEC` into the value passed to `F_SETFL`.

**#41 is closely related and I want to flag it explicitly, because I believe the comment that
closed it was mistaken.** That issue raised the `F_SETFL`/`F_SETFD` namespace distinction for the
`dup`-ed descriptor, and it was resolved on this observation:

> I had missed a very important line in the existing code: `OwnedFd::set_flags` already sets
> `O_CLOEXEC` at the same time as setting `O_NONBLOCK`! So the code already does what you suggest,
> and apart from the small window between the `dup` and `fcntl` calls there is no problem after all.

That reading is what this issue is about. `set_flags` does *mention* `O_CLOEXEC`, but it passes it
to `F_SETFL`, which manipulates file **status** flags and ignores bits outside its documented set --
so the flag is discarded and close-on-exec is never set. The line that appeared to resolve #41 is
the same line reported here as not doing what it appears to do.

I am not asking to reopen #41; its `dup`-window concern is separate and may well be moot. The point
is only that "set_flags already handles O_CLOEXEC" should not be relied on, because it does not.

No severity is claimed. There is no unsoundness -- the `unsafe` block is correct and only the
documented effect is absent -- and the crate's own iterator path never reaches `set_flags`, because
it builds its pair with `UnixStream::pair()` and takes the `WakeMethod::Send` branch. The realistic
cost is a caller who passes a real pipe, follows the documentation, and inherits the write end
across `exec`.

3x3xX3N0N:
`set_flags` passed `O_CLOEXEC` to `F_SETFL`, which only handles file *status* flags and ignores it — so close-on-exec was never set, and the descriptor survives `exec` contrary to the doc comment. Set it via `F_GETFD`/`F_SETFD` with `FD_CLOEXEC` instead, keeping `F_SETFL` for `O_NONBLOCK` only.

Fixes #206

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/low_level/pipe.rs
+++ b/src/low_level/pipe.rs
@@ -107,8 +107,18 @@ impl WakeFd {
             if flags == -1 {
                 return Err(Error::last_os_error());
             }
-            let flags = flags | libc::O_NONBLOCK | libc::O_CLOEXEC;
+            let flags = flags | libc::O_NONBLOCK;
             if libc::fcntl(self.fd.as_raw_fd(), libc::F_SETFL, flags) == -1 {
+                return Err(Error::last_os_error());
+            }
+            // Close-on-exec is a file DESCRIPTOR flag, reached through F_SETFD.
+            // F_SETFL above manipulates file STATUS flags and ignores O_CLOEXEC.
+            let fd_flags = libc::fcntl(self.fd.as_raw_fd(), libc::F_GETFD);
+            if fd_flags == -1 {
+                return Err(Error::last_os_error());
+            }
+            let fd_flags = fd_flags | libc::FD_CLOEXEC;
+            if libc::fcntl(self.fd.as_raw_fd(), libc::F_SETFD, fd_flags) == -1 {
                 return Err(Error::last_os_error());
             }
         }


3x3xX3N0N:
Closing this PR; withdrawing the contribution. Apologies for the noise.

PR #210, opened 2026-08-03
low_level: fix bit arithmetic in android-32 prepare_sigset

### LINKED ISSUE #207: `prepare_sigset` confuses bytes with bits, panicking inside a signal handler on 32-bit Android
## Summary

`emulate_default_handler` is documented async-signal-safe and is designed to be called from
inside a signal handler. On `target_os = "android"` with `target_pointer_width = "32"` it
reaches a hand-rolled `prepare_sigset` that uses `mem::size_of` (a **byte** count) where a
**bit** count is required. Two consequences follow from the one mistake:

1. `assert!` fires for every `Term`-kind signal numbered **5 or above** — including `SIGABRT`,
   `SIGFPE`, `SIGSEGV`, `SIGALRM`, `SIGUSR1`, `SIGUSR2`, `SIGPIPE` and `SIGTERM`. A panic
   inside a signal handler is a panic across an FFI boundary, i.e. undefined behaviour.
2. Even for the signals that pass the assert, the bit selected is the wrong one, so the
   `sigprocmask` unblock targets the wrong signal.

## The code

`src/low_level/signal_details.rs:200-210` (0.3.18):

```rust
                    #[cfg(all(target_os = "android", target_pointer_width = "32"))]
                    unsafe fn prepare_sigset(set: *mut sigset_t, mut signal: c_int) {
                        signal -= 1;
                        let set_raw: *mut libc::c_ulong = set.cast();
                        let size = mem::size_of::<libc::c_ulong>();
                        assert_eq!(set_raw as usize % mem::align_of::<libc::c_ulong>(), 0);
                        let pos = signal as usize / size;
                        assert!(pos < mem::size_of::<sigset_t>() / size);
                        let bit = 1 << (signal as usize % size);
                        set_raw.add(pos).write(bit);
                    }
```

`size` is `size_of::<c_ulong>()`, which is **4** on this target — a count of bytes. It is then
used as the number of signals per word, in both the word index (`signal / size`) and the bit
position (`1 << (signal % size)`). The correct divisor is `8 * size_of::<c_ulong>()`, i.e. 32.

## Worked consequence

On `armv7-linux-androideabi`, libc 0.2.189 defines `pub type sigset_t = c_ulong`
(`src/unix/linux_like/android/b32/mod.rs:8`), so both `size_of::<sigset_t>()` and
`size_of::<c_ulong>()` are 4, and the assert bound `size_of::<sigset_t>() / size` is **1**.

| signal | `signal - 1` | `pos = (signal-1)/4` | `assert!(pos < 1)` |
|---|---|---|---|
| SIGHUP (1) | 0 | 0 | ok |
| SIGINT (2) | 1 | 0 | ok |
| SIGQUIT (3) | 2 | 0 | ok |
| SIGILL (4) | 3 | 0 | ok |
| SIGTRAP (5) | 4 | 1 | **panic** |
| SIGABRT (6) | 5 | 1 | **panic** |
| SIGTERM (15) | 14 | 3 | **panic** |

Only signals 1..=4 survive, and they survive by accident: with the correct divisor of 32 every
signal in 1..=32 lands in word 0, which is what a single-word `sigset_t` requires.

The bit index is wrong in the same way. For `SIGTERM`, `1 << (14 % 4)` is `1 << 2` = 4, whereas
the correct `1 << (14 % 32)` is `1 << 14` = 16384. So where the assert does not fire, the wrong
signal is unblocked.

## Reachability

`prepare_sigset` is called at `signal_details.rs:218`, inside the `DefaultKind::Term` arm of
`emulate_default_handler` (`signal_details.rs:182-224`). That function is public
(`low_level::emulate_default_handler`) and is also reached from the public
`flag::register_conditional_default` (`flag.rs:217-229`), whose registered action body is
`let _ = low_level::emulate_default_handler(signal);` at `flag.rs:225` — and that action runs
**inside the signal handler**.

`pub mod flag;` at `lib.rs:351` carries no `cfg`, so this path exists in every build.

`emulate_default_handler`'s own documentation states, at `signal_details.rs:139`, that
"It is also async-signal-safe."

## Impact

Confined to 32-bit Android, but unconditional there for the common signals. A program using
`register_conditional_default` (the documented pattern for "run the default action on the
second Ctrl-C") panics inside its own signal handler on `SIGTERM`.

Not reachable on any target warden builds; warden compiles this crate only on
`x86_64-unknown-linux-gnu`, where the `#[cfg(not(all(target_os = "android", target_pointer_width = "32")))]`
branch (`signal_details.rs:212-216`) uses `libc::sigemptyset` / `libc::sigaddset` and is correct.

## Suggested fix

```rust
                        let bits = 8 * mem::size_of::<libc::c_ulong>();
                        let pos = signal as usize / bits;
                        assert!(pos < mem::size_of::<sigset_t>() * 8 / bits);
                        let bit = 1 << (signal as usize % bits);
```

The surrounding comment (`signal_details.rs:189-199`) explains that this branch exists because
some Android versions lack `sigemptyset`/`sigaddset` and says the implementation is "inspired by
a libc implementation" and hoped to be correct. It is not.

## How this was found

Line-by-line read of signal-hook 0.3.18 during a `cargo vet` certification of warden's
dependency graph, then confirmed by an adversarial review pass that const-evaluated the
arithmetic on `armv7-linux-androideabi`. The values above were re-derived independently from
the source and from libc's android `b32` type definitions.

Not verified by execution on a live Android device.

## Upstream context

Checked against the default branch before filing: `src/low_level/signal_details.rs:206-215` is
unchanged and still computes `let size = mem::size_of::<libc::c_ulong>();` then uses it as a bit
count in both `signal / size` and `1 << (signal % size)`.

3x3xX3N0N:
The hand-rolled `prepare_sigset` for 32-bit android divided the signal number by `size_of::<c_ulong>()` (bytes, 4) where bits per word (32) is required. Signals 5..=32 tripped the `pos` assert — a panic inside a signal handler when reached via `flag::register_conditional_default` — and for signals 1..=4 the wrong bit was set, unblocking the wrong signal. Compile-checked with `cargo check --target armv7-linux-androideabi`.

Fixes #207

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/low_level/signal_details.rs
+++ b/src/low_level/signal_details.rs
@@ -207,11 +207,12 @@ pub fn emulate_default_handler(signal: c_int) -> Result<(), Error> {
                     unsafe fn prepare_sigset(set: *mut sigset_t, mut signal: c_int) {
                         signal -= 1;
                         let set_raw: *mut libc::c_ulong = set.cast();
-                        let size = mem::size_of::<libc::c_ulong>();
+                        // Words hold bits-per-word signals each, not bytes-per-word.
+                        let bits = 8 * mem::size_of::<libc::c_ulong>();
                         assert_eq!(set_raw as usize % mem::align_of::<libc::c_ulong>(), 0);
-                        let pos = signal as usize / size;
-                        assert!(pos < mem::size_of::<sigset_t>() / size);
-                        let bit = 1 << (signal as usize % size);
+                        let pos = signal as usize / bits;
+                        assert!(pos < mem::size_of::<sigset_t>() / mem::size_of::<libc::c_ulong>());
+                        let bit = 1 << (signal as usize % bits);
                         set_raw.add(pos).write(bit);
                     }
 


3x3xX3N0N:
Closing this PR; withdrawing the contribution. Apologies for the noise.

